PR #11, opened 2020-08-03
disable sens-o-matic by default, allow specifying endpoint via config

alexisprince1994:
**Thanks for contributing to the Prefect Server!**

- [x] add a changelog entry in the `changes/` directory (if appropriate)

As far as I can tell, the bunches of service tests that are failing were failing on the master branch as well.

## What does this PR change?
This PR allows the configuration of the `sens-o-matic` external integration. This disables the external facing calls by default, and when enabled, requires users to specify their own `url` that will receive the payloads.

## Why is this PR important?
This PR will ensure that not all delete events from every users server installation will bombard Prefect's internal sens-o-matic tool, and also allows users to specify their own.

### DIFF
--- a/changes/pr11.yml
+++ b/changes/pr11.yml
@@ -1,0 +1,6 @@
+enhancement:
+    - "Make sens-o-matic configurable - [#11](https://github.com/PrefectHQ/server/pull/11)"
+  
+  contributor:
+    - "[Alex Cano](https://github.com/alexisprince1994)"
+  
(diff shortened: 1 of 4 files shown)


dylanbhughes:
Hey @alexisprince1994! I'm going to close this PR because I'm opening one to remove this whole mechanism for now. We're not reusing code on the cloud side so it's no loss.

PR #13, opened 2020-08-03
Add CI jobs for image builds

joshmeek:
Currently set to build `alpha` tagged images on merge to master


.circleci/config.yml
```diff
@@ -28,20 +28,6 @@ aliases:
         mv /tmp/docker/* /usr/bin
 
 commands:
-  emit_deployment_event:
```
cicdw, line 31: Cool with me but wanted to make sure this was intentional?

joshmeek: Yes, this deployment event does not exist in this repo

PR #31, opened 2020-08-07
Fix Lazarus

cicdw:

## What does this PR change?
This PR fixes a bug in the Lazarus query that relied on a heartbeat being set for Submitted states.  I think this call was lost in the server migration.

## Why is this PR important?
So that Lazarus functions properly.

### DIFF
--- a/tests/api/test_states.py
+++ b/tests/api/test_states.py
@@ -255,7 +255,7 @@ class TestFlowRunStates:
             if not s().is_running() and not s().is_submitted()
         ],
     )
-    async def test_non_running_state_does_not_set_heartbeat(self, state, flow_run_id):
+    async def test_state_does_not_set_heartbeat_unless_running_or_submitted(self, state, flow_run_id):
         flow_run = await models.FlowRun.where(id=flow_run_id).first({"heartbeat"})
         assert flow_run.heartbeat is None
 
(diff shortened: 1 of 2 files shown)


tests/api/test_states.py
```diff
@@ -244,22 +247,35 @@ async def test_trigger_failed_state_does_not_set_end_time(self, flow_run_id):
...
-    async def test_non_running_state_does_not_set_heartbeat(self, flow_run_id):
+    @pytest.mark.parametrize(
+        "state",
+        [
+            s()
+            for s in State.children()
+            if not s().is_running() and not s().is_submitted()
+        ],
+    )
+    async def test_non_running_state_does_not_set_heartbeat(self, state, flow_run_id):
```
jlowin, line 258:
```suggestion
    async def test_state_does_not_set_heartbeat_unless_running_or_submitted(self, state, flow_run_id):
```

tests/api/test_states.py
```diff
@@ -68,11 +68,14 @@ async def test_set_task_run_state_fails_with_wrong_task_run_id(self, state):
                 task_run_id=str(uuid.uuid4()), state=state
             )
 
-    async def test_non_running_state_does_not_set_heartbeat(self, task_run_id):
+    @pytest.mark.parametrize(
+        "state", [s() for s in State.children() if not s().is_running()]
+    )
+    async def test_non_running_state_does_not_set_heartbeat(self, state, task_run_id):
```
jlowin, line 74:
```suggestion
    async def test_state_does_not_set_heartbeat_unless_running(self, state, task_run_id):
```

PR #35, opened 2020-08-10
Remove unused state_id foreign key

jlowin:

## What does this PR change?
This PR removes a foreign key on `task_run.state_id` and `flow_run.state_id` that once pointed to the current state; it is unused and creates performance issues.

### DIFF
--- a/services/postgres/alembic/env.py
+++ b/services/postgres/alembic/env.py
@@ -67,9 +67,7 @@ def apply_hasura_metadata(context):
 
     fn = context.get_context().opts.get("fn")
     if fn and fn.__name__ == "upgrade":
-        dest_rev = context.get_context().opts.get("destination_rev")
-        head = context.script.get_current_head()
-        if dest_rev == "heads" or script.revision == head:
+        if script.revision == context.script.get_current_head():
             hasura_revision = None
         else:
             hasura_revision = script.up_revision
--- a/services/postgres/alembic/env.py
+++ b/services/postgres/alembic/env.py
@@ -70,7 +70,7 @@ def apply_hasura_metadata(context):
         if script.revision == context.script.get_current_head():
             hasura_revision = None
         else:
-            hasura_revision = script.up_revision
+            hasura_revision = script.revision
     elif fn and fn.__name__ == "downgrade":
         hasura_revision = script.down_revision
     else:
--- a/services/postgres/alembic/env.py
+++ b/services/postgres/alembic/env.py
@@ -67,12 +67,14 @@ def apply_hasura_metadata(context):
 
     fn = context.get_context().opts.get("fn")
     if fn and fn.__name__ == "upgrade":
-        if context.get_context().opts.get("destination_rev") == "heads":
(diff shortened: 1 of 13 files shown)


services/postgres/alembic/env.py
```diff
@@ -67,12 +67,12 @@ def apply_hasura_metadata(context):
...
     if fn and fn.__name__ == "upgrade":
-        if context.get_context().opts.get("destination_rev") == "heads":
+        if script.revision == context.script.get_current_head():
             hasura_revision = None
         else:
-            hasura_revision = script.down_revision
+            hasura_revision = script.up_revision
     elif fn and fn.__name__ == "downgrade":
-        hasura_revision = script.revision
+        hasura_revision = script.down_revision
```
cicdw, line 75: why are you having to touch this file?

services/postgres/alembic/versions/2020-08-08T205817_remove_state_id_foreign_keys.py
```diff
@@ -0,0 +1,256 @@
...
+        DROP FUNCTION update_task_run_with_latest_state CASCADE;
+        """
+    )
+    op.execute(
+        """
+        CREATE FUNCTION update_flow_run_with_latest_state() RETURNS trigger
+            LANGUAGE plpgsql
+            AS $$
+                    BEGIN
+                        UPDATE flow_run
```
cicdw, line 34: is this trigger going to run every time a task run sends a heartbeat?

services/postgres/alembic/versions/2020-08-08T205817_remove_state_id_foreign_keys.py
```diff
@@ -0,0 +1,256 @@
...
+                                task_run.version IS NULL
+                                OR (
+                                    new_latest_state.version >= COALESCE(task_run.version, -1)
+                                    AND new_latest_state.timestamp > COALESCE(task_run.state_timestamp, '2000-01-01')));
+                        RETURN NEW;
+                    END;
+                    $$;
+
+            CREATE TRIGGER update_flow_run_after_inserting_state
+            AFTER INSERT ON flow_run_state
```
cicdw, line 114: Ah - if I'm reading this correctly the trigger only runs after a new state is placed in this table, yea?

jlowin: That's correct, this is the trigger that populates the `flow_run` table with the info from the latest flow run state. It has to be modified here to no longer update the `state_id`

services/postgres/alembic/env.py
jlowin: There's a small bug in this code (two actually) which is exposed by destructive migrations like this one -- essentially that it only runs properly if you pass the string `"heads"` as the alembic target. Our unit tests for migrations don't pass `"heads"`; they pass an explicit number of migrations to run (`"+2"`). Those didn't trip this logic correctly and attempted to apply a hasura metadata file that was not compatible with the database state.

PR #43, opened 2020-08-14
Quiet loop logs; improve loop intervals

jlowin:

## What does this PR change?
This PR:
- Quiets debug logs from loop services to heartbeat no more than once every 5 minutes. For rapid services, the noise from heartbeating every loop makes working with logs difficult
- Improves the loop interval calculation so it starts from the beginning of each period not the end. If a service takes one hour to run and has a loop interval of 24 hours, this means it will now start every 24 hours instead of every 25 hours.
- Adds a warning if the loop interval is shorter than the execution time (which would lead to too-rapid execution without pauses)


src/prefect_server/services/loop_service.py
```diff
@@ -49,20 +52,39 @@ async def run(self) -> None:
...
+                )
+
+            # don't log more than once every 5 minutes
+            if pendulum.now("UTC") - last_log > timedelta(minutes=5):
+                self.logger.debug(
+                    f"Heartbeat from {self.name}: next run at {next_run.replace(microsecond=0)}"
+                )
+                last_log = pendulum.now("UTC")
+
+            await asyncio.sleep((next_run - pendulum.now("UTC")).total_seconds())
```
cicdw, line 87: Can you add a test for a loop service whose `run_once` consistently takes longer than `loop_seconds_default`?

jlowin:
@cicdw made the following changes:
- added a `stop()` method to facilitate testing and interactive use (otherwise the services run infinitely)
- added tests, including for the situation you described
- removed the random starting delay - overengineered ✅ 
- turned on asyncio debug mode

cicdw:
Closing this as we've made (most of) these adjustments elsewhere.

jlowin:
Reopening following offline conversation that the remaining enhancements in this PR may be desirable

src/prefect_server/services/loop_service.py
```diff
@@ -40,21 +44,19 @@ def __init__(self, loop_seconds: Union[float, int] = None):
         self.loop_seconds = float(loop_seconds)
         self.name = type(self).__name__
         self.logger = utilities.logging.get_logger(self.name)
+        self._stop_running = False
```
cicdw, line 47: we already have an `_is_running` attribute that you should probably reuse instead

cicdw (changes requested):
Minor request to reuse an existing attribute, otherwise this LGTM

cicdw (approved):
LGTM - this will be a nice QOL improvement

PR #48, opened 2020-08-18
Remove updated and created columns from log table

jlowin:
## Summary

Remove updated/created columns from log table

## Importance

These columns are unnecessary due to the timestamp columns

## Checklist

This PR:


jlowin:
Update: these actually are necessary

PR #58, opened 2020-08-26
Agent Persistence

joshmeek:
## Summary
This _draft_ PR adds the foundation for being able to manage agents in the backend API. The idea here is that users will be able to create an `Agent` and then that agent will have a configuration on it where every time that user wants to use it they can spin up an `AgentInstance` with that configuration. **Note**: agent instances can still be spun up without setting a configuration and they will be persisted. The concept of prefect Agents doesn't really change and this separation of Agent / Instance is more for backend management.

Scenario:
- User enters UI and clicks "New Agent" on the agent management page
- User enters any agent configuration they want and that agent is created
- A command such as `prefect agent start --agent-id XXXX` is displayed that can be used to start an instance of that agent any time it is desired
- User starts an instance of that agent and all flow runs that are submitted by that agent are marked by which agent instance submitted it
- Now a more full agent experience is unlocked where visibility into the agent processes is exposed
- In the case where a user wants to change their agent configuration on the fly they can update it (e.g. via UI) and the agent instance(s) will auto pull the updated configuration
- This will also help identify distressed agent instances

This is meant to be used in collaboration with the [agent_persistence branch in core](https://github.com/PrefectHQ/prefect/compare/agent_persistence). UI side to follow soon!

Opening as a draft PR to get comments and feedback on a few questions I have as well as things like the data models. 

## Changes

Two new tables are added to the DB `agent` which manages the configuration and `agent_instance` which manages things like name, labels, core_version, etc. (standard agent configuration kwargs). This also adds a new field to the flow run table `agent_instance` that is updated by the id of an agent instance when the flow run transitions into a submitted state. Some subsequent mutations are also implemented.

Open questions:
- Currently the `agent_instance` table tracks a `last_query_time` that is updated on every call to retrieve runs. Do we want this?
- Should the `agent` table expose the configuration name as config (current), settings, metadata, etc.
- Should the `agent` table have a `name` field? I can see a scenario of wanting to create an agent w/ a name.

## Importance
Adding the ability to track/manage agents will open the doors to a better experience and provide more detailed information on what is happening across the platform. This also lays the foundation for future agent command patterns because we will be able to direct commands to specific agents/instances.

## Checklist

This PR:

I don't expect tests to pass yet.


src/prefect_server/api/states.py
```diff
@@ -129,6 +130,12 @@ async def set_flow_run_state(
     if state.is_running() or state.is_submitted():
         await api.runs.update_flow_run_heartbeat(flow_run_id=flow_run_id)
 
+    # Set agent instance ID on flow run when submitted by agent
+    if state.is_submitted():
```
cicdw, line 134:
```suggestion
    if state.is_submitted() and agent_instance_id:
```
no reason to hit the DB if no ID was provided

src/prefect_server/utilities/email.py
```diff
@@ -0,0 +1,51 @@
+from typing import List
```
cicdw, line 1: remove this

cicdw (changes requested):

cicdw (changes requested):
Let's also replace all instances of `models.Agent` with `prefect.api.models.Agent` to ensure Cloud compatibility

services/postgres/alembic/versions/2020-08-18T072234_add_agent_persistence.py
```diff
@@ -0,0 +1,109 @@
...
+            sa.ForeignKey("tenant.id", ondelete="CASCADE"),
+            nullable=False,
+            index=True,
+        ),
+        sa.Column("name", sa.String),
+        sa.Column("config", JSONB, nullable=False, server_default="{}"),
+    )
+
+    op.execute(
+        f"""
```
jlowin, line 51:
```suggestion
        """
```

services/postgres/alembic/versions/2020-08-18T072234_add_agent_persistence.py
```diff
@@ -0,0 +1,109 @@
...
+        ),
+        sa.Column(
+            "updated",
+            sa.TIMESTAMP(timezone=True),
+            nullable=False,
+            server_default=sa.func.now(),
+        ),
+        sa.Column(
+            "tenant_id",
+            UUID(),
```
jlowin, line 41: I'm not sure it matters, but this can be left uninstantiated

```suggestion
            UUID,
```

services/postgres/alembic/versions/2020-08-18T072234_add_agent_persistence.py
```diff
@@ -0,0 +1,109 @@
...
+        EXECUTE PROCEDURE set_updated_timestamp();
+        """
+    )
+
+    op.create_table(
+        "agent_instance",
+        sa.Column(
+            "id", UUID, primary_key=True, server_default=sa.func.gen_random_uuid()
+        ),
+        sa.Column(
```
jlowin, line 64: Should this table have an `updated` column?

services/postgres/alembic/versions/2020-08-18T072234_add_agent_persistence.py
```diff
@@ -0,0 +1,109 @@
...
+        sa.Column("labels", JSONB, nullable=False, server_default="[]"),
+        sa.Column(
+            "last_queried",
+            sa.TIMESTAMP(timezone=True),
+            nullable=True,
+            server_default=sa.func.now(),
+        ),
+    )
+
+    op.add_column(
```
jlowin, line 96: If we're anticipating showing all runs from a certain agent, this column might need an index

src/prefect_server/graphql/agents.py
```diff
@@ -0,0 +1,61 @@
...
+from prefect_server.utilities import context
+from prefect_server.utilities.graphql import mutation
+
+
+@mutation.field("register_agent_instance")
+async def resolve_register_agent_instance(
+    obj: Any, info: GraphQLResolveInfo, input: dict
+) -> dict:
+    agent_id = await api.agents.register_agent_instance(
+        tenant_id=input.get("tenant_id"),
```
jlowin, line 15:
```suggestion
        tenant_id=input["tenant_id"],
```

src/prefect_server/graphql/agents.py
```diff
@@ -0,0 +1,61 @@
...
+        labels=input.get("labels", []) or [],
```
jlowin, line 16: These are required fields in the GQL schema, so you can rely on them having expected types
```suggestion
        labels=input["labels"],
```

src/prefect_server/graphql/schema/agents.graphql
```diff
@@ -0,0 +1,76 @@
...
+
+input register_agent_instance_input {
+  "the tenant"
+  tenant_id: UUID!
+  "instance labels"
+  labels: [String]!
+  "agent id"
+  agent_id: UUID
+  "instance type"
+  type: String
```
jlowin, line 35: You access this as `input["type"]` in the resolver, so it should be required

```suggestion
  type: String!
```

src/prefect_server/graphql/schema/agents.graphql
```diff
@@ -0,0 +1,76 @@
...
+
+# ------------------------------------------------------
+# agent inputs
+# ------------------------------------------------------
+
+input register_agent_instance_input {
+  "the tenant"
+  tenant_id: UUID!
+  "instance labels"
+  labels: [String]!
```
jlowin, line 31: This will prevent `[null]` from being passed
```suggestion
  labels: [String!]!
```

src/prefect_server/graphql/schema/agents.graphql
```diff
@@ -0,0 +1,76 @@
...
+input delete_agent_input {
+  "The ID of the agent to delete"
+  agent_id: UUID!
+}
+
+input set_agent_config_input {
+  "The ID of the agent to update"
+  agent_id: UUID!
+  "A config to set for this agent"
+  config: JSON
```
jlowin, line 63: This needs to be required, because 1) it is accessed as `input['config']` in the resolver and 2) null configs are not allowed in the db

```suggestion
  config: JSON!
```

jlowin (changes requested):

services/postgres/alembic/versions/2020-08-18T072234_add_agent_persistence.py
joshmeek: We actually decided to remove is in favor of the last_queried column because other than that this shouldn't really be updated

services/postgres/alembic/versions/2020-08-18T072234_add_agent_persistence.py
joshmeek: Initially the idea was that the flow run display would show the agent instance that submitted it and not the other way around. I see what you're saying though

src/prefect_server/api/runs.py
```diff
@@ -344,9 +344,33 @@ async def delete_flow_run(flow_run_id: str) -> bool:
     return bool(result.affected_rows)  # type: ignore
 
 
+@register_api("runs.update_flow_run_agent_instance")
+async def update_flow_run_agent_instance(
```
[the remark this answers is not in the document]
joshmeek: This was another thing we discussed but it (along with the no index on the agent_instance_id on the flow run table) was decided to keep on the flow run table because in the information we are choosing to display a user is only going to want to see the agent instance which kicked off the most recent run

joshmeek:
> Let's also replace all instances of `models.Agent` with `prefect.api.models.Agent` to ensure Cloud compatibility

@cicdw What do you mean? The `Agent` model is in `prefect_server.models`

cicdw:
@joshmeek as an example, if we re-use the `delete_agent` route in Cloud, we want to make sure that `models.Agent().delete()` uses the _Cloud_ model and not the Server model.  Analogously to the `prefect.api` calls, we use `prefect.api.models.Agent` to ensure that the correct model is used in each call.

src/prefect_server/api/runs.py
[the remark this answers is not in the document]
cicdw: Yea, I view the "multiple agents running one flow" pattern as a future enhancement, and in the meantime we can definitely store the agent instance ID on the Submitted state directly - in fact we can add the agent ID to the "Submitted" log for visibility

src/prefect_server/api/runs.py
```diff
@@ -344,9 +344,33 @@ async def delete_flow_run(flow_run_id: str) -> bool:
...
+    flow_run_id: str, agent_instance_id: str
+) -> None:
+    """
+    Updates the agent instance of a flow run
+
+    Args:
+        - flow_run_id (str): the flow run id
+        - agent_instance_id (str): the id of an agent instance submitting the flow run
```
cicdw, line 356: gotta change these instance ID references :/

src/prefect_server/database/_models.py
```diff
@@ -328,6 +329,33 @@ class Message(HasuraModel):
...
+@plugins.register_model("AgentConfig")
+class AgentConfig(HasuraModel):
+    __hasura_type__ = "agent_config"
+
+    id: UUIDString = None
+    created: datetime.datetime = None
+    updated: datetime.datetime = None
+    tenant_id: UUIDString = None
+    name: str = None
+    config: dict = None
```
cicdw, line 341: let's change this column to "settings"

src/prefect_server/graphql/schema/agents.graphql
```diff
@@ -0,0 +1,76 @@
...
+input delete_agent_config_input {
+  "The ID of the agent config to delete"
+  agent_id: UUID!
+}
+
+input set_agent_config_input {
+  "The ID of the agent config to update"
+  agent_config_id: UUID!
+  "A config to set for this agent config"
+  config: JSON!
```
jlowin, line 63:
```suggestion
  settings: JSON!
```

src/prefect_server/graphql/schema/agents.graphql
```diff
@@ -0,0 +1,76 @@
...
+  delete_agent(input: delete_agent_input!): success_payload
+
+  "Create a new agent config"
+  create_agent_config(input: create_agent_config_input!): agent_config_id_payload
+
+  "Delete an agent config"
+  delete_agent_config(input: delete_agent_config_input!): success_payload
+
+  "Update an agent config's config"
+  set_agent_config(input: set_agent_config_input!): success_payload
```
jlowin, line 19: If we have multiple atomic mutations (`set_agent_config_settings`, `set_agent_config_name`, `set_agent_config_notification_preferences`) then I like "set" here; if this will eventually become one large mutation I prefer "update" instead: `update_agent_config`

src/prefect_server/api/states.py
```diff
@@ -129,6 +130,10 @@ async def set_flow_run_state(
     if state.is_running() or state.is_submitted():
         await api.runs.update_flow_run_heartbeat(flow_run_id=flow_run_id)
 
+    # Set agent ID on flow run when submitted by agent
+    if state.is_submitted() and agent_id:
+        await api.runs.update_flow_run_agent(flow_run_id=flow_run_id, agent_id=agent_id)
```
jlowin, line 135: If the agent ID is invalid for any reason (including being valid and then having the agent deleted from the db), it will error and abort the state setting. Perhaps `update_flow_run_agent` should return `False` on error, instead of raising ValueError?

joshmeek: Updated but FWIW the heartbeat update functions the same way. Not sure if that should change also

src/prefect_server/api/agents.py
```diff
@@ -0,0 +1,142 @@
...
+        - bool: whether the update was successful
+    """
+    if not agent_config_id:
+        raise ValueError("Invalid agent config ID.")
+
+    update = {}
+    if name:
+        update["name"] = name
+    if settings:
+        update["settings"] = settings
```
cicdw, line 139: This sort of logic actually prevents users from being able to, e.g., clear their settings (because an empty dict or `None` is never respected as the input).  We could do `if settings is not None:` instead, and "clearing" settings simply means providing an empty dict to `settings`

jlowin (approved):
🚀

cicdw (approved):

cicdw (approved):

cicdw (approved):
Awesome work @joshmeek ! Thanks for your patience through this extended review process

PR #63, opened 2020-08-31
Deprecate `api.flows` settings that are actually for flow groups

jlowin:
## Summary

Deprecates `api.flows` functions that actually modify flow group settings

## Changes

Routes for enabling/disabling heartbeats, lazarus, and version locking are now available as `api.flow_groups.*` routes. The old `api.flows.*` routes remain available as deprecated routes.

This is because the introduction of flow groups has changed how flow settings are managed, but these routes predate the change and are in the wrong module.

## Checklist

This PR:


src/prefect_server/api/deprecated/flows.py
```diff
@@ -0,0 +1,172 @@
+from prefect import api
+from prefect_server.database import models
+from prefect.utilities.plugins import register_api
+
+
+@register_api("flows._update_flow_setting")
+async def _update_flow_setting(flow_id: str, key: str, value: any) -> models.FlowGroup:
```
cicdw, line 7: Not important since these are deprecated anyway, but why is this "private"?

src/prefect_server/api/flow_groups.py
```diff
@@ -7,6 +7,42 @@
 from prefect.utilities.plugins import register_api
 
 
+@register_api("flow_groups._update_setting")
+async def _update_setting(flow_group_id: str, key: str, value: any) -> None:
```
cicdw, line 11: Same question as above

cicdw (dismissed):
LGTM for the changes, but I did have a question

src/prefect_server/api/deprecated/flows.py
jlowin: Historic artifact from an earlier attempt at the server API. I will update the new version in `/api/flow_groups` (but leave this one for compat)

PR #65, opened 2020-09-01
Fix flow run creation not using default flow group parameters

### LINKED ISSUE #64: QUICK RUN cannot take newer Default Parameter updated from UI
## Description

The QUICK RUN button will take default Parameter which is given when registered, but it still takes this default Parameter no matter how I modify it on UI later (settings->Parameters->Default Parameters).

## Expected Behavior

"QUICK RUN" should take the parameter after user update it on UI. Like the code below, when I register and run this flow, it will return 3 (2+1), but when I modify the parameter from 2 to 50, it still return 3.

## Reproduction

```
from prefect import task, Flow, Parameter
import prefect

logger = prefect.utilities.logging.get_logger()
@task
def print_plus_one(x):
    print(x + 1)
    logger.warning(x+1)

with Flow('Parameterized Flow') as flow:
    x = Parameter('x', default = 2)
    print_plus_one(x=x)

flow.register(project_name="Test Project")
```

## Environment

Prefect version: 0.13.2

joshmeek:
## Summary
Closes #64 
Flow run creation was not using the default parameters set on the flow group 

## Changes
Updates the parameter value passed to the flow run creation (previously it was unused)

## Checklist

This PR:

### DIFF
--- a/changes/issue64.yaml
+++ b/changes/issue64.yaml
@@ -1,2 +1,2 @@
 fix:
-  - "Fix flow run creation not using default parameters from flow group - [#64](https://github.com/PrefectHQ/server/issues/64"
+  - "Fix flow run creation not using default parameters from flow group - [#64](https://github.com/PrefectHQ/server/issues/64)"
--- a/changes/issue64.yaml
+++ b/changes/issue64.yaml
@@ -1,2 +1,0 @@
-fix:
-  - "Fix flow run creation not using default parameters from flow group - [#64](https://github.com/PrefectHQ/server/issues/64"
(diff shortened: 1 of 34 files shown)


changes/issue64.yaml
```diff
@@ -0,0 +1,2 @@
+fix:
+  - "Fix flow run creation not using default parameters from flow group - [#64](https://github.com/PrefectHQ/server/issues/64"
```
cicdw, line 2:
```suggestion
  - "Fix flow run creation not using default parameters from flow group - [#64](https://github.com/PrefectHQ/server/issues/64)"
```

PR #66, opened 2020-09-01
Remove black dependency pin and reformat files

joshmeek:
This removes the black pin and reformats files. This pin was causing issues in development because `prefect` is up to date on versions.

cc @cicdw Why was this pinned?

### DIFF
--- a/src/prefect_server/api/cloud_hooks.py
+++ b/src/prefect_server/api/cloud_hooks.py
@@ -76,12 +76,16 @@ async def create_cloud_hook(
         if set(config or {}) != {"url"}:
             raise ValueError("Invalid config; expected `url`")
     elif type == "TWILIO":
-        if set(config or {}) != {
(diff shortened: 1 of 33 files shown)


src/prefect_server/api/cloud_hooks.py
```diff
@@ -76,12 +76,16 @@ async def create_cloud_hook(
...
+        if (
+            set(config or {})
+            != {
+                "account_sid",
+                "auth_token",
+                "messaging_service_sid",
+                "to",
+            }
+            or not config.get("to")
+        ):
```
cicdw, line 88: @joshmeek reformatted code like this ^^^ is why it was pinned; this looks awful

joshmeek: I actually think it looks better! THE FUTURE IS NOW

PR #76, opened 2020-09-03
Apollo Dockerfile Refactoring and Improvements.

gabrielcalderon:
## Summary
* Reduce overall size of final docker image by using less layers.
* Changed base image from **node:10.15.0** to **node:10.15.0-slim**

## Changes
* Overall image size of latest tag uncompressed was reduced from **1.03 GB** to **271 MB**, a _decrease_ of **73%**.
* Added **PREFECT_SERVER_VERSION** build argument to Dockerfile. Defaults to **master**
* Added **NODE_VERSION** build argument to Dockerfile. Defaults to **10.15.0**
* Added image labels.
* Both **PREFECT_VERSION** and **PREFECT_SERVER_VERSION** now default to "master".

## Future work
* The current CI process should be updated to pass the **Prefect Server Version** using **PREFECT_SERVER_VERSION** to the docker build. The current Dockerfile defaults to **master** for both Prefect, Prefect Server, and Apollo, unless an argument is passed.

## Importance
* Reduce network usage when downloading image.
* Reduce storage usage when using Prefect Docker Images.
* New Dockerfile is similar to PrefectHQ/Prefect.

## Checklist

This PR:

- [- ] adds new tests (if appropriate)
- [x ] adds a change file in the `changes/` directory (if appropriate)


.circleci/config.yml
```diff
@@ -202,7 +202,7 @@ workflows:
           docker-username: DOCKER_HUB_USER
           image: 'prefecthq/server'
           path: '.'
-          extra_build_args: '--build-arg PREFECT_VERSION=master --build-arg RELEASE_TIMESTAMP=$(date -u +"%Y-%m-%dT%H:%M:%SZ")'
+          extra_build_args: '--build-arg PREFECT_VERSION=master --build-arg BUILD_DATE=$(date -u +"%Y-%m-%dT%H:%M:%SZ")'
```
cicdw, line 205: `RELEASE_TIMESTAMP` is a setting that is actually used within the application, so we need to keep it

services/apollo/Dockerfile
```diff
@@ -1,15 +1,30 @@
-FROM node:10.15.0
-ARG VERSION
-ENV PREFECT_SERVER_VERSION=${VERSION}
+ARG NODE_VERSION=${NODE_VERSION:12}
```
cicdw, line 1: We'll need to do some heavy QA on our side to make sure this upgrade is non-breaking, so we should stick to `10.15.0` for now and open an issue for upgrading to `12`

cicdw (changes requested):
Two changes that will need to be reverted to ensure no breaking changes, but otherwise this looks good!

services/apollo/Dockerfile
gabrielcalderon: @cicdw There's currently 3 point releases for 10.15, does it have to be specifically 10.15.0 or would 10.15.3 work?

gabrielcalderon:
@cicdw Requested changes were made.

services/apollo/Dockerfile
```diff
@@ -1,15 +1,30 @@
...
+LABEL org.label-schema.url="https://www.prefect.io/"
+LABEL org.label-schema.version=${PREFECT_VERSION}
+LABEL org.label-schema.build-date=${RELEASE_TIMESTAMP}
 
 WORKDIR /apollo
-COPY post-start.sh .
-RUN chmod +x post-start.sh
+COPY . .
+
+RUN npm install && \
```
cicdw, line 26: Is there a reason you changed from `npm ci` to `npm install`?  I think we should stick with `npm ci` to ensure no surprising changes make their way into the image

gabrielcalderon: Agree, it should be ```npm ci```

cicdw (approved):
This is really great work @gabrielcalderon - thank you!

PR #77, opened 2020-09-03
Introduce API for setting flow run / task run names

### LINKED ISSUE #75: Add routes for renaming flow runs and tasks
## Current behavior

Currently, flow runs can be named at creation only or given random names, and task runs are randomly named.

## Proposed behavior

A new GQL mutation allows runs to be renamed. By incorporating this into a Prefect task, users could rename runs based on parameters, schedules, inputs, or anything else. Prefect runs on Prefect™️

jlowin:
## Summary

Allows flow run and task run names to be customized

## Changes

Introduces two new GraphQL routes: `set_flow_run_name` and `set_task_run_name`

## Importance

Closes #75 

## Checklist

This PR:


services/postgres/alembic/versions/2020-09-03T095523_add_task_run_name.py
```diff
@@ -0,0 +1,26 @@
...
+
+# revision identifiers, used by Alembic.
+revision = "e148cf9f1e5b"
+down_revision = "70528cee0d2b"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.add_column("task_run", sa.Column("name", sa.String))
```
joshmeek, line 22: What's the need for a task_run name? IIRC this existed a while ago but was removed because it was an unnecessary piece of information

jlowin: We removed randomly-generated ones that we used to have, but there have been user requests for setting task run names, for example for dynamically generated mapped children.

joshmeek: For the dynamically generated mapped children do you mean users are going to come in and name each of those task runs individually or adopt something like https://github.com/PrefectHQ/prefect/pull/2974 (which doesn't yet have a nice implementation)?

cicdw: The idea would be that if they provided a custom callable for producing a name, at the beginning of each run we could call this API route for them and update the name - so yea, the hope is that we can update that PR to simply call this route at the appropriate time

cicdw (commented):
Other than the migration conflicts this LGTM!

cicdw (approved):

PR #78, opened 2020-09-03
Use "up revision" to name hasura metadata files

jlowin:
## Summary

Adjusts hasura metadata files to use the latest revision ID instead of the prior revision ID

## Changes

This PR seems small but has large implications for concurrent migration PRs.

Consider creating a new database migration when the current alembic revision ID is `abc123`. Today, a new hasura metadata archive is created as `metadata-abc123.yaml`. This means that two open PRs, both including migrations, will also both have a `metadata-abc123.yaml` file. Since Hasura metadata is particularly opaque, this leads to difficult-to-debug merge conflicts and most devs will resolve it by simply starting over.

This PR modifies the metadata archive to use the new alembic revision ID. So if the current alembic revision ID is `abc123`, a PR might generate a new revision `xyz456` and corresponding `metadata-xyz456.yaml`. A concurrent PR could generate a new revision `mno789` and corresponding `metadata-mno789.yaml`. 

Critically, there is still a merge conflict to be resolved, as the linear migration path will have to be resolved. However, because there are two different files, instead of one conflicted file, the conflict can be resolved more easily and we have unit tests that are very helpful in determining if it exists.

So while this PR doesn't magically solve merge conflicts for concurrent migrations, it does introduce the possibility that resolving them is as simple as tweaking the last PR's `down_revision`, without needing to regenerate the hasura metadata.

## Checklist

This PR:


cicdw (commented):
A few questions:

1.) What determines what metadata file to apply when performing a downgrade?  Does this naming change mess that up in any way?
2.) I don't think I see how this solves the ultimate problem - you'd still have to recreate the hasura metadata file to take into account whatever schema changes occurred from the conflicting migration changes, no?  In fact, this new process seems like it would _obscure_ that fact and make it easier to introduce a bad metadata version.

jlowin:
@cicdw for question 1 [this is the line](https://github.com/PrefectHQ/server/pull/78/files#diff-ee5e66680e2b759dc364593c5ad4a44cR70) that determines the revision to pull, it's been updated to reflect the new logic.

For question 2, the primary issue is that metadata files are created using the "down" revision ID. This means two different PRs that branch from the same origin will each create a file with that name, resulting in obscure merge conflicts and necessitating the regeneration of the hasura metadata (itself complex because the user needs to remember to roll back to the original metadata state of their PR). Basically, two different PRs that generate two different IDs for their migrations are nonetheless using the same (down) ID to represent the associated hasura metadata - makes no sense.

With this change, the metadata files are created using the "up" revision ID. This means two different PRs working off the same branch will create two different files and avoid a conflict. It's true that if the metadata changes, users will need to work to ensure its up to date but its much more straightforward to do so without the confusion of the merge conflict, and any errors will be flagged by an existing unit test.

cicdw (approved):

PR #87, opened 2020-09-11
Update return payload of mapped_children in the presence of a bad data

cicdw:
This ensures the return type matches the behavior in Cloud for UI compatibility


jlowin (dismissed):

cicdw:
Closing because I realize now that my code replicates the current behavior with an additional (and unnecessary) DB hit.

PR #89, opened 2020-09-14
Use prefect.models in favor of database.models

joshmeek:
## Summary

Attempting to access the models via
```python
from prefect import api

api.models.Agent
```

Raises a `box.BoxKeyError: "'Box' object has no attribute 'models'"` and instead it needs to be `from prefect import models`.

## Changes
Updates all API files that called `prefect_server.database.models` in favor of `prefect.models`

## Importance
Code reuse and ensuring smooth repo-specific functionality

## Checklist

This PR:


cicdw (approved):
This is great - thanks for identifying and fixing @joshmeek !

PR #90, opened 2020-09-15
81 flow concurrency limits

### LINKED ISSUE #81: Flow Concurrency Limits
## Current behavior

Currently, Flow Concurrency Limits are a cloud-only feature. There was a [PR](https://github.com/PrefectHQ/prefect/pull/2382) against Core's server while it was still a part of the core package, but was dropped due to the migration of Server to it's own repo and sharing the core code with Cloud.

## Proposed behavior

The implementation of Flow Concurrency Limits should allow labeling of Flows and limiting concurrent executions of flows.

A few notable questions / changes from the previous PR:
* The database table requires a `tenant_id` column foreign keyed to the `tenant` table.
* The database table should not include the `description` column.
* The publicly available mutations should be `update_flow_concurrency_limit` (which acts as an upsert) and `delete_flow_concurrency_limit`

## Example

See Cloud's documentation of Flow Concurrency Limits :)

alexisprince1994:
## Summary

Closes #81 
This PR adds Flow Concurrency Limits to Server, which should mirror Cloud's implementation.

## Changes

This PR adds GraphQL mutations to update and delete flow concurrency limits, as well as adds checks for actually limiting concurrent flow runs via the state setting pipeline.

## Importance

This PR will help bring the Server codebase up to parity with Cloud's feature set.

## Checklist

This PR:


alexisprince1994:
After resolving conflicts w/ master, I noticed the code style tests failing. After running `black` on the codebase locally, the test still seems to fail, but as a result of the merge another ~25 files were reformatted. So now the PR contains a bunch of touched files :persevere:. Not sure how this happened and also not sure why the code style CI job is still failing.

alexisprince1994:
Looks like the failing tests are from the migrations & hasura metadata. I'll take a look at these over the next couple of days as it's a result from the merge conflict.

alexisprince1994:
Upsides and downsides: Database migration test that was previously failing because of merge has been fixed. All consequential tests are now passing.

Downside: It appears the code style test (and CI hook) is failing. I'm running python 3.7.4 locally and am using the same configuration files as the repo. Running `black --check .` results in:
```
(.venv) ➜  server git:(81-flow-concurrency-limits) black --check .  
All done! ✨ 🍰 ✨
140 files would be left unchanged.
(.venv) ➜  server git:(81-flow-concurrency-limits) python --version
Python 3.7.4
```

joshmeek:
Hi @alexisprince1994  it looks like your local git may not be authenticated with your GitHub account! Generally you can see this when your commits show up as the grey octocat logo.

If you want to resolve this (and attribute the commits to yourself) you could try options like this:
- https://docs.github.com/en/enterprise/2.18/user/github/committing-changes-to-your-project/why-are-my-commits-linked-to-the-wrong-user#commits-are-not-linked-to-any-user
- https://docs.github.com/en/github/setting-up-and-managing-your-github-user-account/setting-your-commit-email-address#setting-your-commit-email-address-in-git

I took this PR for a spin and locally I adjusted a few things in the schema to match the adaptation in Cloud (variable names and such). If you want I can push my changes to this branch or I could comment them here, no preference.

One thing I encountered when running this was I would run into an issue when a flow was queued due to no available slots and then the task run attempts to enter a running state while the flow is Queued and it raises:

```
[2020-09-28 18:15:04] ERROR - prefect.CloudTaskRunner | Failed to set task state with error: ClientError([{'message': 'State update failed for task run ID 8badfeea-2827-49f3-b2fa-a9dd1ce10d59: provided a running state but associated flow run cb22b62d-2318-4e22-9207-7d9ae833c4f4 is not in a running state.', 'locations': [{'line': 2, 'column': 5}], 'path': ['set_task_run_states'], 'extensions': {'code': 'INTERNAL_SERVER_ERROR', 'exception': {'message': 'State update failed for task run ID 8badfeea-2827-49f3-b2fa-a9dd1ce10d59: provided a running state but associated flow run cb22b62d-2318-4e22-9207-7d9ae833c4f4 is not in a running state.'}}}])
Traceback (most recent call last):
  File "/Users/josh/Desktop/code/prefect/src/prefect/engine/cloud/task_runner.py", line 128, in call_runner_target_handlers
    cache_for=self.task.cache_for,
  File "/Users/josh/Desktop/code/prefect/src/prefect/client/client.py", line 1321, in set_task_run_state
    version=version,
  File "/Users/josh/Desktop/code/prefect/src/prefect/client/client.py", line 294, in graphql
    raise ClientError(result["errors"])
prefect.utilities.exceptions.ClientError: [{'message': 'State update failed for task run ID 8badfeea-2827-49f3-b2fa-a9dd1ce10d59: provided a running state but associated flow run cb22b62d-2318-4e22-9207-7d9ae833c4f4 is not in a running state.', 'locations': [{'line': 2, 'column': 5}], 'path': ['set_task_run_states'], 'extensions': {'code': 'INTERNAL_SERVER_ERROR', 'exception': {'message': 'State update failed for task run ID 8badfeea-2827-49f3-b2fa-a9dd1ce10d59: provided a running state but associated flow run cb22b62d-2318-4e22-9207-7d9ae833c4f4 is not in a running state.'}}}]
```

Thinking this may be due to something in the `get_runs_in_queue` logic as I am able to see both runs getting picked up (output below). Not sure if the intended behavior is that only N runs will be pulled from the queue to match the amount of available concurrency slots or we expect all runs to be grabbed and then in the flow process they wait until they can proceed 🤔  going to cc @cicdw for his thoughts on this concept. Other than that this is looking sharp!

```
[2020-09-28 18:14:32,009] DEBUG - agent | Querying flow run metadata
[2020-09-28 18:14:32,038] INFO - agent | Found 1 flow run(s) to submit for execution.
[2020-09-28 18:14:32,039] DEBUG - agent | Next query for flow runs in 0.25 seconds
[2020-09-28 18:14:32,039] DEBUG - agent | Updating states for flow run 7882b2d8-8f05-44d2-bca0-0e0cda47839b
[2020-09-28 18:14:32,041] DEBUG - agent | Flow run 7882b2d8-8f05-44d2-bca0-0e0cda47839b is in a Scheduled state, updating to Submitted
[2020-09-28 18:14:32,143] INFO - agent | Deploying flow run 7882b2d8-8f05-44d2-bca0-0e0cda47839b
[2020-09-28 18:14:32,148] DEBUG - agent | Submitted flow run 7882b2d8-8f05-44d2-bca0-0e0cda47839b to process PID 12009
[2020-09-28 18:14:32,176] DEBUG - agent | Completed flow run submission (id: 7882b2d8-8f05-44d2-bca0-0e0cda47839b)
[2020-09-28 18:14:32,294] DEBUG - agent | Querying for flow runs
[2020-09-28 18:14:32,369] DEBUG - agent | Found flow runs ['cb22b62d-2318-4e22-9207-7d9ae833c4f4']
[2020-09-28 18:14:32,369] DEBUG - agent | Querying flow run metadata
[2020-09-28 18:14:32,398] INFO - agent | Found 1 flow run(s) to submit for execution.
[2020-09-28 18:14:32,398] DEBUG - agent | Updating states for flow run cb22b62d-2318-4e22-9207-7d9ae833c4f4
[2020-09-28 18:14:32,401] DEBUG - agent | Flow run cb22b62d-2318-4e22-9207-7d9ae833c4f4 is in a Scheduled state, updating to Submitted
[2020-09-28 18:14:32,403] DEBUG - agent | Next query for flow runs in 0.25 seconds
[2020-09-28 18:14:32,488] INFO - agent | Deploying flow run cb22b62d-2318-4e22-9207-7d9ae833c4f4
[2020-09-28 18:14:32,492] DEBUG - agent | Submitted flow run cb22b62d-2318-4e22-9207-7d9ae833c4f4 to process PID 12010
[2020-09-28 18:14:32,524] DEBUG - agent | Completed flow run submission (id: cb22b62d-2318-4e22-9207-7d9ae833c4f4)
```

alexisprince1994:
@joshmeek Thanks so much for the in-depth review! I always forget that I need to change my default credentials when working on new github repos (default uses my gitlab credentials). I'm planning on doing that over the next couple of days!

I'd love it if you could just push the changes to have them align with how things look in Cloud, especially based on your description of how they are mostly semantic differences.

So the intended behavior (based on previous conversations, might not be up to date) if there are 2 runs and only enough slots for 1 of them is to only return 1 from the `get_runs_in_queue` function to avoid deadlocking agents / execution infrastructure. It sounds like the situation you produced regarding the queueing is a bug; I'll write some more tests and try to figure out how / why that happened.

joshmeek:
In the scenario where I schedule a flow, it begins running, and then while it's running another flow attempts to run this works great. However in the scenario where I schedule two flows to run, with a single concurrency slot, and then start an agent they run into an issue:

```
[2020-10-12 21:30:20,982] INFO - agent | Deploying flow run 7052e5af-fb8c-49ac-b4ba-7178e70af2af
[2020-10-12 21:30:20,984] INFO - agent | Deploying flow run cbb2021f-ac7b-4b14-878f-09294f5f16ef
[2020-10-12 21:30:22] INFO - prefect.CloudFlowRunner | Beginning Flow run for 'ETL'
[2020-10-12 21:30:22] INFO - prefect.CloudFlowRunner | Beginning Flow run for 'ETL'
[2020-10-12 21:30:22] DEBUG - prefect.CloudFlowRunner | Using executor type LocalExecutor
[2020-10-12 21:30:22] DEBUG - prefect.CloudFlowRunner | Using executor type LocalExecutor
[2020-10-12 21:30:22] DEBUG - prefect.CloudFlowRunner | Flow 'ETL': Handling state change from Scheduled to Running
[2020-10-12 21:30:22] DEBUG - prefect.CloudFlowRunner | Flow 'ETL': Handling state change from Scheduled to Running
[2020-10-12 21:30:22] DEBUG - prefect.CloudFlowRunner | Flow run is in a Queued state. Sleeping for at most 30.00 seconds and attempting to run again.
[2020-10-12 21:30:22] DEBUG - prefect.CloudFlowRunner | Flow run is in a Queued state. Sleeping for at most 30.00 seconds and attempting to run again.
```

What I believe is happening is the `get_runs_in_queue` isn't successfully blocking/securing the concurrency slot for the runs. I'm thinking that this is going to need a field somewhere that can be incremented/decremented in order to secure this slot because the get runs in queue is running into a race condition. I'm not sure if the database is the best place to put that although it seems like the only option.

alexisprince1994:
@joshmeek This does look like a bug. The desired behavior in this scenario is that both flow runs get returned to the agent (since there are 0 runs occupying slots at poll time). Then, one of 2 things should occur: Ideally, when the agent submits work, one of the runs should succeed and enter the `Submitted` state and the other should fail to transition, but there can be a race condition here where they're both entering the `Submitted` state. If the race condition does occur, when the runs attempt to enter the `Running` state, the first that attempts to do so should be put into a `Queued` state by server. The second flow run should be able to successfully transition to `Running` from `Submitted`, since the `Queued` flow run is no longer occupying a concurrency slot.

I'm going to try to isolate how this occurred, because what you're seeing is not the intended behavior.

joshmeek:
@alexisprince1994 here's another thing I encountered (in the case where one run actually starts running):

Order of events:
- Set two to scheduled
- Start agent
- Two are pulled
- Two are submitted
- On start one enters running, one enters queued
- Running flow runs until completion fine
- The queued flow was locally queued for 30 seconds but in server it seems like it's queued for 10 minutes. The first task starts to change from pending to running but can’t because the flow run is in a Queued state

Logs from when it happened:
```

[2020-10-12 22:01:07,708] INFO - agent | Starting LocalAgent with labels ['test', 'Joshs-MacBook-Pro-2.local', 'azure-flow-storage', 'gcs-flow-storage', 's3-flow-storage', 'github-flow-storage', 'webhook-flow-storage', 'gitlab-flow-storage']
[2020-10-12 22:01:07,708] INFO - agent | Agent documentation can be found at https://docs.prefect.io/orchestration/
[2020-10-12 22:01:07,708] INFO - agent | Agent connecting to the Prefect API at http://localhost:4200
[2020-10-12 22:01:07,717] INFO - agent | Waiting for flow runs...
[2020-10-12 22:01:07,813] INFO - agent | Found 2 flow run(s) to submit for execution.
[2020-10-12 22:01:07,943] INFO - agent | Deploying flow run b7fcbf5d-27cb-416f-8c36-c3145ff03c31
[2020-10-12 22:01:07,986] INFO - agent | Deploying flow run 97625370-f3b9-4461-8dde-fea8d2f40c86
[2020-10-12 22:01:08] INFO - prefect.CloudFlowRunner | Beginning Flow run for 'ETL'
[2020-10-12 22:01:08] DEBUG - prefect.CloudFlowRunner | Using executor type LocalExecutor
[2020-10-12 22:01:08] INFO - prefect.CloudFlowRunner | Beginning Flow run for 'ETL'
[2020-10-12 22:01:08] DEBUG - prefect.CloudFlowRunner | Using executor type LocalExecutor
[2020-10-12 22:01:08] DEBUG - prefect.CloudFlowRunner | Flow 'ETL': Handling state change from Scheduled to Running
[2020-10-12 22:01:09] DEBUG - prefect.CloudFlowRunner | Flow 'ETL': Handling state change from Scheduled to Running
[2020-10-12 22:01:09] DEBUG - prefect.CloudFlowRunner | Flow run is in a Queued state. Sleeping for at most 30.00 seconds and attempting to run again.
[2020-10-12 22:01:09] INFO - prefect.CloudTaskRunner | Task 'extract': Starting task run...
[2020-10-12 22:01:09] DEBUG - prefect.CloudTaskRunner | Task 'extract': Handling state change from Pending to Running
[2020-10-12 22:01:09] DEBUG - prefect.CloudTaskRunner | Task 'extract': Calling task.run() method...
[2020-10-12 22:01:19] DEBUG - prefect.LocalResult | Starting to upload result to /Users/josh/.prefect/results/prefect-result-2020-10-12t22-01-19-317679-00-00...
[2020-10-12 22:01:19] DEBUG - prefect.LocalResult | Finished uploading result to /Users/josh/.prefect/results/prefect-result-2020-10-12t22-01-19-317679-00-00...
[2020-10-12 22:01:19] DEBUG - prefect.CloudTaskRunner | Task 'extract': Handling state change from Running to Success
[2020-10-12 22:01:19] INFO - prefect.CloudTaskRunner | Task 'extract': finished task run for task with final state: 'Success'
[2020-10-12 22:01:19] INFO - prefect.CloudTaskRunner | Task 'transform': Starting task run...
[2020-10-12 22:01:19] DEBUG - prefect.CloudTaskRunner | Task 'transform': Handling state change from Pending to Running
[2020-10-12 22:01:19] DEBUG - prefect.CloudTaskRunner | Task 'transform': Calling task.run() method...
[2020-10-12 22:01:24] DEBUG - prefect.CloudFlowRunner | Checking flow run state...
[2020-10-12 22:01:29] DEBUG - prefect.LocalResult | Starting to upload result to /Users/josh/.prefect/results/prefect-result-2020-10-12t22-01-29-558025-00-00...
[2020-10-12 22:01:29] DEBUG - prefect.LocalResult | Finished uploading result to /Users/josh/.prefect/results/prefect-result-2020-10-12t22-01-29-558025-00-00...
[2020-10-12 22:01:29] DEBUG - prefect.CloudTaskRunner | Task 'transform': Handling state change from Running to Success
[2020-10-12 22:01:29] INFO - prefect.CloudTaskRunner | Task 'transform': finished task run for task with final state: 'Success'
[2020-10-12 22:01:29] INFO - prefect.CloudTaskRunner | Task 'load': Starting task run...
[2020-10-12 22:01:29] DEBUG - prefect.CloudTaskRunner | Task 'load': Handling state change from Pending to Running
[2020-10-12 22:01:29] DEBUG - prefect.CloudTaskRunner | Task 'load': Calling task.run() method...
[2020-10-12 22:01:39] INFO - prefect.CloudFlowRunner | Beginning Flow run for 'ETL'
[2020-10-12 22:01:39] DEBUG - prefect.CloudFlowRunner | Using executor type LocalExecutor
[2020-10-12 22:01:39] DEBUG - prefect.CloudFlowRunner | Checking flow run state...
[2020-10-12 22:01:39] INFO - prefect.CloudTaskRunner | Task 'extract': Starting task run...
[2020-10-12 22:01:39] DEBUG - prefect.CloudTaskRunner | Task 'extract': Handling state change from Pending to Running
[2020-10-12 22:01:39] ERROR - prefect.CloudTaskRunner | Failed to set task state with error: ClientError([{'message': 'State update failed for task run ID 64d6115b-4cea-4001-ad66-82bf3aa930b6: provided a running state but associated flow run b7fcbf5d-27cb-416f-8c36-c3145ff03c31 is not in a running state.', 'locations': [{'line': 2, 'column': 5}], 'path': ['set_task_run_states'], 'extensions': {'code': 'INTERNAL_SERVER_ERROR', 'exception': {'message': 'State update failed for task run ID 64d6115b-4cea-4001-ad66-82bf3aa930b6: provided a running state but associated flow run b7fcbf5d-27cb-416f-8c36-c3145ff03c31 is not in a running state.'}}}])
Traceback (most recent call last):
  File "/Users/josh/Desktop/code/prefect/src/prefect/engine/cloud/task_runner.py", line 128, in call_runner_target_handlers
    cache_for=self.task.cache_for,
  File "/Users/josh/Desktop/code/prefect/src/prefect/client/client.py", line 1373, in set_task_run_state
    version=version,
  File "/Users/josh/Desktop/code/prefect/src/prefect/client/client.py", line 294, in graphql
    raise ClientError(result["errors"])
prefect.utilities.exceptions.ClientError: [{'message': 'State update failed for task run ID 64d6115b-4cea-4001-ad66-82bf3aa930b6: provided a running state but associated flow run b7fcbf5d-27cb-416f-8c36-c3145ff03c31 is not in a running state.', 'locations': [{'line': 2, 'column': 5}], 'path': ['set_task_run_states'], 'extensions': {'code': 'INTERNAL_SERVER_ERROR', 'exception': {'message': 'State update failed for task run ID 64d6115b-4cea-4001-ad66-82bf3aa930b6: provided a running state but associated flow run b7fcbf5d-27cb-416f-8c36-c3145ff03c31 is not in a running state.'}}}]
[2020-10-12 22:01:39] INFO - prefect.CloudTaskRunner | Task 'extract': finished task run for task with final state: 'ClientFailed'
[2020-10-12 22:01:39] INFO - prefect.CloudTaskRunner | Task 'transform': Starting task run...
[2020-10-12 22:01:39] DEBUG - prefect.CloudTaskRunner | Task 'transform': not all upstream states are finished; ending run.
[2020-10-12 22:01:39] INFO - prefect.CloudTaskRunner | Task 'transform': finished task run for task with final state: 'Pending'
[2020-10-12 22:01:39] INFO - prefect.CloudTaskRunner | Task 'load': Starting task run...
[2020-10-12 22:01:39] DEBUG - prefect.CloudTaskRunner | Task 'load': not all upstream states are finished; ending run.
[2020-10-12 22:01:39] INFO - prefect.CloudTaskRunner | Task 'load': finished task run for task with final state: 'Pending'
[2020-10-12 22:01:39] DEBUG - prefect.CloudFlowRunner | Checking flow run state...
[2020-10-12 22:01:39] WARNING - prefect.CloudFlowRunner | Flow run is no longer in a running state; the current state is: <Queued: "Queued by flow run concurrency limit">
[2020-10-12 22:01:39] INFO - prefect.CloudFlowRunner | Flow run RUNNING: terminal tasks are incomplete.
Here's your data: [10, 20, 30]
[2020-10-12 22:01:39] DEBUG - prefect.CloudTaskRunner | Task 'load': Handling state change from Running to Success
[2020-10-12 22:01:39] INFO - prefect.CloudTaskRunner | Task 'load': finished task run for task with final state: 'Success'
[2020-10-12 22:01:39] DEBUG - prefect.CloudFlowRunner | Checking flow run state...
[2020-10-12 22:01:39] INFO - prefect.CloudFlowRunner | Flow run SUCCESS: all reference tasks succeeded
[2020-10-12 22:01:39] DEBUG - prefect.CloudFlowRunner | Flow 'ETL': Handling state change from Running to Success
```

joelluijmes:
Hey @alexisprince1994 are you still working on this PR? If you want I could try to chime in, this seems like something I'm looking for 😄

alexisprince1994:
@joelluijmes The PR (as far as I know) is complete, however I got swamped at my day job so I haven't had enough time to actually spin up a server and throw a bunch of scenarios at it. I believe I put the fix in place to ensure what @joshmeek described above no longer happens, but can't say for certain.

If you'd like to help get this pushed through, going through testing scenarios would be incredibly helpful!

joelluijmes:
Okay, just got it running and tested a similar scenario: 
1. Don't start any agents
2. Setup a label which limits concurrency to limit=1
3. Create 10 jobs (which basically just wait 30 seconds).
4. Start an agent
5. See what happens:

All flows went in scheduled state, then two jobs failed, at some point two jobs were running simultaneously and the other jobs executed serially. Also it seems a job hanged for a minute.

**Flow runs**
```json
{
  "data": {
    "flow_run": [
      {
        "id": "edc0afc8-eca5-4650-97a1-bf72ad348cde",
        "name": "curious-spaniel",
        "scheduled_start_time": "2020-11-13T11:45:28.310275+00:00",
        "start_time": null,
        "end_time": "2020-11-13T11:49:03.746061+00:00",
        "state": "Failed"
      },
      {
        "id": "8b10f8b3-5c7b-4572-9613-27ede9daa6e2",
        "name": "pompous-polecat",
        "scheduled_start_time": "2020-11-13T11:45:29.76821+00:00",
        "start_time": null,
        "end_time": "2020-11-13T11:49:03.742416+00:00",
        "state": "Failed"
      },
      {
        "id": "636f4342-2108-4144-bd8b-561b2a8d0611",
        "name": "vague-lion",
        "scheduled_start_time": "2020-11-13T11:45:30.488854+00:00",
        "start_time": "2020-11-13T11:49:39.088623+00:00",
        "end_time": "2020-11-13T11:50:09.525326+00:00",
        "state": "Success"
      },
      {
        "id": "3bc10880-ec53-4dd5-bfaf-261e1773f1a9",
        "name": "shaggy-hyrax",
        "scheduled_start_time": "2020-11-13T11:45:25.364819+00:00",
        "start_time": "2020-11-13T11:50:09.626311+00:00",
        "end_time": "2020-11-13T11:50:40.114355+00:00",
        "state": "Success"
      },
      {
        "id": "05989913-3603-4e56-ae22-fc4c47682ddd",
        "name": "jovial-cockle",
        "scheduled_start_time": "2020-11-13T11:45:23.889184+00:00",
        "start_time": "2020-11-13T11:51:09.771514+00:00",
        "end_time": "2020-11-13T11:51:40.362441+00:00",
        "state": "Success"
      },
      {
        "id": "f7af9c11-17e7-4c3b-943a-228573ad47aa",
        "name": "sincere-coot",
        "scheduled_start_time": "2020-11-13T11:45:29.032027+00:00",
        "start_time": "2020-11-13T11:51:09.772512+00:00",
        "end_time": "2020-11-13T11:51:40.359273+00:00",
        "state": "Success"
      },
      {
        "id": "e7a785ac-31cb-4e5b-87ce-243ad3a30b54",
        "name": "solid-pogona",
        "scheduled_start_time": "2020-11-13T11:45:26.087557+00:00",
        "start_time": "2020-11-13T11:51:40.421773+00:00",
        "end_time": "2020-11-13T11:52:10.820671+00:00",
        "state": "Success"
      },
      {
        "id": "62e9f6b7-5a5f-4100-8542-f3127b4d2609",
        "name": "mindful-rat",
        "scheduled_start_time": "2020-11-13T11:45:26.823989+00:00",
        "start_time": "2020-11-13T11:52:40.340499+00:00",
        "end_time": "2020-11-13T11:53:11.332282+00:00",
        "state": "Success"
      },
      {
        "id": "215109a7-6797-4f59-a7ee-ac3feddeb686",
        "name": "unnatural-shellfish",
        "scheduled_start_time": "2020-11-13T11:45:27.576607+00:00",
        "start_time": "2020-11-13T11:53:40.820704+00:00",
        "end_time": "2020-11-13T11:54:11.198824+00:00",
        "state": "Success"
      },
      {
        "id": "64591a40-a54a-40bc-b435-896deb548c74",
        "name": "tasteful-hornet",
        "scheduled_start_time": "2020-11-13T11:45:24.628759+00:00",
        "start_time": "2020-11-13T11:54:41.211313+00:00",
        "end_time": "2020-11-13T11:55:11.634704+00:00",
        "state": "Success"
      }
    ]
  }
}
```

**Agent logs**
```
[2020-11-13 11:49:02,815] INFO - agent | Starting DockerAgent with labels ['semaphore-test']
[2020-11-13 11:49:02,815] INFO - agent | Agent documentation can be found at https://docs.prefect.io/orchestration/
[2020-11-13 11:49:02,815] INFO - agent | Agent connecting to the Prefect API at http://localhost:4200
[2020-11-13 11:49:02,821] INFO - agent | Waiting for flow runs...
[2020-11-13 11:49:03,101] INFO - agent | Found 10 flow run(s) to submit for execution.
[2020-11-13 11:49:03,445] ERROR - agent | Logging platform error for flow run edc0afc8-eca5-4650-97a1-bf72ad348cde
[2020-11-13 11:49:03,446] ERROR - agent | Logging platform error for flow run 8b10f8b3-5c7b-4572-9613-27ede9daa6e2
[2020-11-13 11:49:03,451] INFO - agent | Deploying flow run 64591a40-a54a-40bc-b435-896deb548c74
[2020-11-13 11:49:03,453] INFO - agent | Pulling image eu.gcr.io/gynzy-intern/prefect/sleepy:2020-11-13t11-43-37-166560-00-00...
[2020-11-13 11:49:03,477] INFO - agent | Deploying flow run 05989913-3603-4e56-ae22-fc4c47682ddd
[2020-11-13 11:49:03,479] INFO - agent | Pulling image eu.gcr.io/gynzy-intern/prefect/sleepy:2020-11-13t11-43-37-166560-00-00...
[2020-11-13 11:49:03,482] INFO - agent | Deploying flow run 62e9f6b7-5a5f-4100-8542-f3127b4d2609
[2020-11-13 11:49:03,484] INFO - agent | Pulling image eu.gcr.io/gynzy-intern/prefect/sleepy:2020-11-13t11-43-37-166560-00-00...
[2020-11-13 11:49:03,534] INFO - agent | Deploying flow run e7a785ac-31cb-4e5b-87ce-243ad3a30b54
[2020-11-13 11:49:03,536] INFO - agent | Pulling image eu.gcr.io/gynzy-intern/prefect/sleepy:2020-11-13t11-43-37-166560-00-00...
[2020-11-13 11:49:03,539] INFO - agent | Deploying flow run 3bc10880-ec53-4dd5-bfaf-261e1773f1a9
[2020-11-13 11:49:03,541] INFO - agent | Pulling image eu.gcr.io/gynzy-intern/prefect/sleepy:2020-11-13t11-43-37-166560-00-00...
[2020-11-13 11:49:03,572] INFO - agent | Deploying flow run 215109a7-6797-4f59-a7ee-ac3feddeb686
[2020-11-13 11:49:03,574] INFO - agent | Pulling image eu.gcr.io/gynzy-intern/prefect/sleepy:2020-11-13t11-43-37-166560-00-00...
[2020-11-13 11:49:03,588] INFO - agent | Deploying flow run 636f4342-2108-4144-bd8b-561b2a8d0611
[2020-11-13 11:49:03,591] INFO - agent | Pulling image eu.gcr.io/gynzy-intern/prefect/sleepy:2020-11-13t11-43-37-166560-00-00...
[2020-11-13 11:49:03,607] INFO - agent | Deploying flow run f7af9c11-17e7-4c3b-943a-228573ad47aa
[2020-11-13 11:49:03,610] INFO - agent | Pulling image eu.gcr.io/gynzy-intern/prefect/sleepy:2020-11-13t11-43-37-166560-00-00...
[2020-11-13 11:49:03,787] ERROR - agent | Error while deploying flow: ClientError([{'message': 'Unable to get flow run concurrency slot. Aborting.', 'locations': [{'line': 2, 'column': 5}], 'path': ['set_flow_run_states'], 'extensions': {'code': 'INTERNAL_SERVER_ERROR', 'exception': {'message': 'Unable to get flow run concurrency slot. Aborting.'}}}])
[2020-11-13 11:49:03,789] ERROR - agent | Error while deploying flow: ClientError([{'message': 'Unable to get flow run concurrency slot. Aborting.', 'locations': [{'line': 2, 'column': 5}], 'path': ['set_flow_run_states'], 'extensions': {'code': 'INTERNAL_SERVER_ERROR', 'exception': {'message': 'Unable to get flow run concurrency slot. Aborting.'}}}])
WARNING: Could not store access token in cache: database is locked
[2020-11-13 11:49:06,235] INFO - agent | Successfully pulled image eu.gcr.io/gynzy-intern/prefect/sleepy:2020-11-13t11-43-37-166560-00-00...
[2020-11-13 11:49:06,236] INFO - agent | Successfully pulled image eu.gcr.io/gynzy-intern/prefect/sleepy:2020-11-13t11-43-37-166560-00-00...
[2020-11-13 11:49:06,239] INFO - agent | Successfully pulled image eu.gcr.io/gynzy-intern/prefect/sleepy:2020-11-13t11-43-37-166560-00-00...
[2020-11-13 11:49:06,239] INFO - agent | Successfully pulled image eu.gcr.io/gynzy-intern/prefect/sleepy:2020-11-13t11-43-37-166560-00-00...
[2020-11-13 11:49:06,240] INFO - agent | Successfully pulled image eu.gcr.io/gynzy-intern/prefect/sleepy:2020-11-13t11-43-37-166560-00-00...
[2020-11-13 11:49:06,348] INFO - agent | Successfully pulled image eu.gcr.io/gynzy-intern/prefect/sleepy:2020-11-13t11-43-37-166560-00-00...
[2020-11-13 11:49:06,349] INFO - agent | Successfully pulled image eu.gcr.io/gynzy-intern/prefect/sleepy:2020-11-13t11-43-37-166560-00-00...
[2020-11-13 11:49:06,354] INFO - agent | Successfully pulled image eu.gcr.io/gynzy-intern/prefect/sleepy:2020-11-13t11-43-37-166560-00-00...
```

Don't have a clue what is wrong yet, but I'll have a look. What I also find interesting (and wasn't expecting), is that all images were pulled immediately, or is that the expected behavior?

alexisprince1994:
@joelluijmes Thanks for taking the time to do this! So to catch you up with the decisions so far that haven't happened on this ticket (and to address what you've seen as behavior):

* There is a known race condition when you have N concurrency slots available, but there are N + 1 flow runs that would use that slot being returned from the `get_runs_in_queue` call, N + 1 flow runs are returned. This behavior allows much more straightforward logic on the backend, and prevents us sending runs we know won't have a slot available to the agent. _However_, this does mean that N + 1 flow runs can move from `Scheduled` to `Submitted`, which (assuming no race conditions), shouldn't happen, however we accept that this is a race condition that will happen and we plan around it.

* Runs that _should_ transition from `Submitted` to `Running`, but do not have an available concurrency slot are placed in the `Queued` state, with a `start_time`of 10 minutes in the future. This interaction means that the CloudFlowRunner process that sends requests to Server will receive a notification that the state is not `Running`, and it should bail out of the current run. However, since the state is `Queued`, the CloudFlowRunner retries every ~30 seconds to run the flow run again. This is likely the behavior that caused the hanging you were seeing.

* The idea of locking was thrown around a bit, and ultimately we decided to go for a "best effort" implementation of concurrency limiting in Server due to the complexity of adding locking that'd work for all cases. Locking would involve adding another table to manage the locks, which we wanted to avoid.

Now, the behavior you are seeing that _isn't_ explained / planned:
* Failed runs: AFAIK, there is no reason that runs should fail with this change. At worst, runs should either be entering a `Queued` state, or attempt and repeatedly fail to be `Submitted`
* 2 runs running concurrently: This shouldn't happen. There are race conditions built into the code, but there are 2 separate race conditions that need to happen (first on the transition from `Scheduled` to `Submitted` and second from `Submitted` to `Running`) in order for this to occur. We may need to either revert back to the more complicated logic where we're counting flow runs to return to the Agent or introduce a jitter into the client. I'm guessing we'd like to avoid jittering the client since the Cloud implementation doesn't require a jitter.

I'll try to make some time this week to look into these, especially the `Failed` runs, as this _really_ shouldn't be happening. If you have some time before then and find the issue, please either let me know and I can add tests around it or add the tests yourself!

joelluijmes:
Thanks for your reply, good to know you intent on finishing this 👍 

> * There is a known race condition when you have N concurrency slots available, but there are N + 1 flow runs that would use that slot being returned from the `get_runs_in_queue` call, N + 1 flow runs are returned. This behavior allows much more straightforward logic on the backend, and prevents us sending runs we know won't have a slot available to the agent. _However_, this does mean that N + 1 flow runs can move from `Scheduled` to `Submitted`, which (assuming no race conditions), shouldn't happen, however we accept that this is a race condition that will happen and we plan around it.

Wouldn't it make sense to let all overflowing flow runs go to Queued? Or is the Submitted transition required first? (I've no idea how the state machine works).

> * Runs that _should_ transition from `Submitted` to `Running`, but do not have an available concurrency slot are placed in the `Queued` state, with a `start_time`of 10 minutes in the future. This interaction means that the CloudFlowRunner process that sends requests to Server will receive a notification that the state is not `Running`, and it should bail out of the current run. However, since the state is `Queued`, the CloudFlowRunner retries every ~30 seconds to run the flow run again. This is likely the behavior that caused the hanging you were seeing.

Does that mean that overflowing runs are delayed by 10 minutes? Or are they will be retried after the 30 second interval?

> * The idea of locking was thrown around a bit, and ultimately we decided to go for a "best effort" implementation of concurrency limiting in Server due to the complexity of adding locking that'd work for all cases. Locking would involve adding another table to manage the locks, which we wanted to avoid.

Ah, an in memory locking mechanism wouldn't suffice in case prefect was running multiple instances I suppose 😅 
Anyhow, wouldn't my behavior be explained due lack of locking? I'm not that into your code, but I would guess that is the root cause. 

I also saw the [following line](https://github.com/PrefectHQ/server/pull/90/files#diff-fa45aa44801328c209aaecd64b28803f7e3db966f54598fec211e2e7d2745672R127):
```python
is_occupying_run_slot = num_records == 1
```

Not sure if can happen, but wouldn't it be safer to check for` >= 1`?


Let me know if you need more logs to find out whats going wrong, I'll try to see if I can reproduce it in a test.

alexisprince1994:
> Wouldn't it make sense to let all overflowing flow runs go to Queued? Or is the Submitted transition required first? (I've no idea how the state machine works).
The state machine (for flow runs, ignoring `Queued` states) goes `Scheduled` -> `Submitted` -> `Running` -> `Finished` (`Success` or `Failed`). By injecting `Queued`, we aim to intercept runs that that end up in the `Submitted` state that are occupying a concurrency slot that is unavailable for them. When we add concurrency limits in, transitioning from `Scheduled` to `Submitted` actually occupies a slot (as you noted in your first comment, when flow runs transition to `Submitted`, external things happen, like pulling docker images), and we check again when transitioning from `Submitted` to `Running` due to the race condition when an agent pulls multiple runs. However, if we know there aren't concurrency slots available, we don't allow the transition from `Scheduled` to `Submitted` (instead of populating them as `Queued`).

> Does that mean that overflowing runs are delayed by 10 minutes? Or are they will be retried after the 30 second interval?
It means that they are retried after the 30 second interval on the process that was running the flow run initially. However, if that process dies for whatever reason, Server knows to give that flow run to an agent after the 10 minutes.

> Ah, an in memory locking mechanism wouldn't suffice in case prefect was running multiple instances I suppose 😅
Correct, we wanted to avoid an in-memory lock for this reason, as well as recovery from a failure.


> Not sure if can happen, but wouldn't it be safer to check for >= 1?
This shouldn't be able to happen, as the query is looking up a flow run by it's ID (which is guaranteed unique). It will only ever return 0 or 1 record. The purpose of this is that if there's a concurrency limit of 1, and a flow run transitions from `Scheduled` to `Submitted`, that concurrency slot is occupied. However, when transitioning from `Submitted` to `Running`, we needed a way to indicate that this run is occupying a slot already, so that we don't see the 0 slots and reject the state transition, when in reality we'd be switching from `Submitted` to `Running`, both of which occupy a slot, so we are _net_ changing the slot usage by 0.

> Let me know if you need more logs to find out whats going wrong, I'll try to see if I can reproduce it in a test.
More logs are always helpful! As noted above, I'm most concerned around the runs that failed, as this PR should never cause runs to fail. At most, it should delay them, allowing them to run successfully down the road.

joelluijmes:
> The state machine (for flow runs, ignoring `Queued` states) goes `Scheduled` -> `Submitted` -> `Running` -> `Finished` (`Success` or `Failed`). By injecting `Queued`, we aim to intercept runs that that end up in the `Submitted` state that are occupying a concurrency slot that is unavailable for them. When we add concurrency limits in, transitioning from `Scheduled` to `Submitted` actually occupies a slot (as you noted in your first comment, when flow runs transition to `Submitted`, external things happen, like pulling docker images), and we check again when transitioning from `Submitted` to `Running` due to the race condition when an agent pulls multiple runs. However, if we know there aren't concurrency slots available, we don't allow the transition from `Scheduled` to `Submitted` (instead of populating them as `Queued`).
 
Ah alright, so that is one of the race conditions currently happening.
 
> Correct, we wanted to avoid an in-memory lock for this reason, as well as recovery from a failure.

Makes sense, I’ve not used it (only heard about it) but postgres also offers a generic locking mechnism: https://www.postgresql.org/docs/12/explicit-locking.html#ADVISORY-LOCKS Maybe an easy solution to adding locks?

alexisprince1994:
After having some time today (finally) to do further testing of this, I was able to dig further into the following things from above and have comments on their larger architectural impact below. All of these instances occurred with the set up described by @joelluijmes:

* Due to the lack of any kind of locking mechanism, the race condition of going from `Submitted` -> `Running` / `Queued` was often not hit. For reference, with the setup above, 6+ flow runs (typically around 8) were getting past both `Scheduled` -> `Submitted` and `Submitted` -> `Running` checks. I was able to put in a temporary solution of adding a jitter around the concurrency check, but this feels hacky and I'd think the more appropriate place to put a jitter would be in the `prefect.Client` calls. However, I'm not sure of the thought process of the Prefect team (cc @joshmeek / @cicdw ) of adding a client side jitter when the Cloud side doesn't require it. This also still doesn't "fix" our race condition, only makes it less likely to occur. I'm okay with leaving the server implementation being a "best effort" implementation, but having our best effort work <= 40% of the time isn't great.

* The `Failed` states noted by Joel from above were as a result of the Agent failing to deploy these flow runs. There currently isn't any mechanism for catching an "expected" failure while getting changed to `Submitted`. I'd think that a special case of matching the exception text on the error received by the Agent would handle this, but again, not sure of the Prefect team's thought on doing that. It'd create a hard to spot dependency between the two codebases. *However*, one important thing to note is that `Lazarus` will eventually revive these runs, _but_ one thing to note is that if enough runs are using the concurrency limits, there _might_ (and this is pretty far fetched) be a case where certain runs are getting marked as failed so often that they never run, even though the only thing causing them to fail are a lack of concurrency slots.


Thoughts team? I noticed that in the current implementation, empty labels being passed to `try_take_flow_concurrency_slots` would immediately return, so in the state setting pipeline I did the following (not wanting to commit the changes yet since I'm not certain of them):

```python
# Previous
if state.is_running() or state.is_submitted():
    can_transition = ...

# New
# Not doing the check if the run doesn't have labels
if (state.is_running() or state.is_submitted()) and flow_run.labels:
    # a guesstimate at a large enough jitter to demonstrate a naive solution at race condition mitigation
    await asyncio.sleep(random.random() * 3)
    can_transition = ...
```

src/prefect_server/api/flow_concurrency_limits.py
```diff
@@ -0,0 +1,162 @@
...
+        raise ValueError("Invalid flow concurrency limit ID.")
+
+    result = await models.FlowConcurrencyLimit.where(id=limit_id).delete()
+    return bool(result.affected_rows)
+
+
+@register_api("flow_concurrency_limits.try_take_flow_concurrency_slots")
+async def try_take_flow_concurrency_slots(
+    tenant_id: str, limit_names: List[str], current_state: Optional[State] = None
+) -> bool:
```
alexisprince1994, line 86: The previous signature took a `flow_run_id: str = None` parameter instead of `current_state: Optional[State] = None`. This was changed as all places this check is done already have queried the flow run and have existing state data available. For compatibility concerns, this can be changed back, but will likely slow down performance.

src/prefect_server/api/runs.py
```diff
@@ -473,14 +477,29 @@ async def get_runs_in_queue(
...
             continue
 
+        # Agent label filtering
         # if the run has no labels but labels were provided, skip
         if not run_labels and labels:
             continue
 
+        # Concurrency label filtering
+        # Concurrency limiting final check is in state setting pipeline
+        if run_labels:
```
alexisprince1994, line 492: Not locking on concurrency checks since this is supposed to be a more liberal filter. The only purpose is to prevent sending agents flow runs that we already know there are 0 available slots for, not to do fancy counting math that sends _only_ the correct number of flow runs (although we can do it, it just increases complexity).

src/prefect_server/api/states.py
```diff
@@ -100,6 +123,101 @@ async def set_flow_run_state(
...
+        # or is attempting to occupy a slot (X -> Submitted)
+        existing_state = state_schema.load(flow_run.serialized_state)
+
+        # Due to the Agent process submitting multiple runs from Scheduled
+        # to Submitted at once, we need to acquire a lock to both check &
+        # set the new state if there are concurrency limits in place to
+        # avoid race conditions between the time a check has occurred and
+        # when an update has occurred.
+        server_context = context.get_context()
+        flow_concurrency_lock = server_context.get(
```
alexisprince1994, line 149: An alternate implementation doesn't need to rely on the `flow_concurrency_lock` or `has_been_concurrency_checked` parameters. See comment in GraphQL code about `flow_concurrency_lock` reason.

`has_been_concurrency_checked` was initially added to avoid needing to manually manage the lock within the method (with the assumption that a differing function signature was not a problem). We can revert the signature back to the cloud compatible one, then manage acquiring the lock before the concurrency check and release it after the state insert. Coincidentally, this would also be better for performance since we'd avoid querying the run info again.

src/prefect_server/api/states.py
```diff
@@ -100,6 +123,101 @@ async def set_flow_run_state(
...
+                    state=state,
+                    version=version,
+                    agent_id=agent_id,
+                    has_been_concurrency_checked=True,
+                )
+
+        if existing_state.is_scheduled() and state.is_submitted():
+            buffer_seconds = random.randint(15, 45)
+
+            state = Queued(
```
alexisprince1994, line 177: This use of `Queued` deviates from how Cloud uses `Queued`. This scenario occurs when an Agent attempts to submit a run and there are no slots left. If we raise an error here, the run will be put into a `Failed` state by the Agent, which isn't what we want.

`buffer_seconds` was a guesstimate of a decent implementation. It would be nice to add a jitter in here to avoid agents getting the same batches of runs over and over though.

src/prefect_server/graphql/extensions.py
```diff
@@ -36,6 +37,22 @@ def log_error(exc: Exception) -> None:
         logger.error({"traceback": traceback.format_exc(), "context": ctx})
 
 
+class PrefectConcurrencyLimiter(Extension):
```
alexisprince1994, line 40: A single process global concurrency lock. I'm not married to this implementation (or naming), but it seemed to fit well with the injection of the `headers` into the context in the `PrefectHeader` extension.

src/prefect_server/graphql/states.py
```diff
@@ -23,6 +23,7 @@ async def resolve_set_flow_run_states(
 
     server_context = context.get_context()
     agent_id = server_context.get("headers", {}).get("x-prefect-agent-id")
+    flow_concurrency_lock = server_context.get("flow_concurrency_lock")
```
alexisprince1994, line 26: We can remove this and have the `api.states.set_flow_run_state` call directly pull this lock from the server context. I had originally added it here as we're also pulling `agent_id` from the context, so I wanted to group those together.

One downside of this (and only passing it as a parameter instead of reading from the server context in the function directly) is that we'd only be able to test the limiting behavior via the `graphql` tests and not the `api` tests.

tests/api/test_states.py
```diff
@@ -492,3 +499,356 @@ async def test_cancelling_non_running_flow_run_returns_cancelled(self, flow_run_
...
+
+class TestQueueFlowRun:
+    async def test_doeesnt_transition_on_submitted_transition_with_no_slots(
+        self,
+        flow_id: str,
+        labeled_flow_run_id: str,
+        flow_concurrency_limit: models.FlowConcurrencyLimit,
+    ):
+        """
+        Cloud compatability test; If a flow run is attempting to transition to
```
alexisprince1994, line 512: This is currently an inaccurate comment, and will change depending on whether we're okay with Server and Cloud slightly deviating in which runs get Queued.

madkinsz:
@alexisprince1994 Hey! I can't make a lot of these decisions as I'm still getting onboarded here at Prefect but I'll try to give this a review soon.

src/prefect_server/api/flow_concurrency_limits.py
```diff
@@ -0,0 +1,162 @@
...
+        - bool: Whether the run already occupies or can occupy a concurrency
+            slot.
+    """
+
+    if not limit_names:
+        # Unlabeled runs always have available slots
+        return True
+
+    if current_state is not None:
+        is_occupying_run_slot = (
```
alexisprince1994, line 118: Technically we only need to check if the state is `Submitted` here, since flow runs in a `Running` state won't ever need to acquire a concurrency slot. _However_, I think leaving the check for `Running` and `Submitted` helps keep the intention clear since we're checking `Running` and `Submitted` later in the function.

src/prefect_server/api/states.py
```diff
@@ -20,7 +23,12 @@
 
 @register_api("states.set_flow_run_state")
 async def set_flow_run_state(
-    flow_run_id: str, state: State, version: int = None, agent_id: str = None
+    flow_run_id: str,
+    state: State,
+    version: int = None,
+    agent_id: str = None,
+    flow_concurrency_lock: Semaphore = None,
```
alexisprince1994, line 30: An alternate implementation doesn't need to rely on the flow_concurrency_lock or has_been_concurrency_checked parameters. See comments later in this file (`has_been_concurrency_checked`) and in the GraphQL mutation code ( `flow_concurrency_lock`).

jacksund:
Thought I would jump in here because @samreay and I have similar issue that relies on flow concurrency. Since my suggest fix is different from what this thread is trying to do, let me know if I should open a separate pull request for this.

We run into issues with Agents crashing from too many concurrent flow runs. His case is with `DockerAgent` and mine with `LocalAgent`. I believe this is because too many subprocesses (or Docker instances) are created, causing my computer to eat up RAM/CPU and kill the parent Agent process.

This pull request looks to bring flow concurrency limits to Agents to fix that problem, but I'd like to propose a fix that is slightly different: 
1. Let Prefect Cloud handle flow-run concurrency for any _**single**_ flow.
2. Let Agents hand flow-run concurrency for **_all_** flows. In other words, Agents have a hard-cap on how many total flow runs they are allow to have running concurrently -- regardless of which flows those runs belong too.

This will set a hard limit on how many flows an Agent can run at a given time and will prevent the Agent from submitting too many subprocesses or spinning up too many Docker instances. 

For `LocalAgent`, we can look at the `agent.processes` property before calling `agent.deploy_flow`. The pre-check before launching the flow could be as simple as...
```
if len(agent.processes) < max_concurrency:
    agent.deploy_flow()
else:
   # should this wait here or do something else? How will this play with Prefect async approach?
```
This fix could be applied on this line https://github.com/PrefectHQ/prefect/blob/849c903404da1c352453c41d67c9508de1959922/src/prefect/agent/agent.py#L414 

Since this in the parent Agent class, we can't actually use the `self.processes` attribute because it's not universal to all subclass Agents. Instead, the base class may need to add a abstract method such as `get_number_of_active_flow_runs()` that each Agent implements.

cicdw:
I have some bittersweet news -- we have decided to close this PR, for a few reasons:
- as this feature already exists in Cloud (and is built on patterns / infra that is not available in Server), we don't believe we have the bandwidth to properly maintain two separate implementations of the same feature at this time
- @madkinsz recently found a way to cleanly implement a _very similar_ (although not identical) feature that we believe will unblock all Server users who have requested this; follow the issue and the idea here: https://github.com/PrefectHQ/server/issues/213

We genuinely appreciate the effort and enthusiasm that went into this work, and look forward to everyone's feedback on the new feature proposal and implementation!

PR #94, opened 2020-09-22
Replace all uses of database.models with prefect.models

joshmeek:


cicdw (approved):
This is great - hopefully this will prevent any future DB bugs from sneaking in!

PR #95, opened 2020-09-23
Match agents to existing entries during registration

joshmeek:
## Summary
If an agent is constantly restarting this could lead to a large amount of data being created because each restart previously created a new instance of an agent row in the db. This PR relieves that by matching agents against tenant_id, name, type, core_version, and labels. If an agent is already found to have been created with that configuration the ID of that agent is returned from registration and a new entry is not created.

Example: Due to auto deploys of some infrastructure this is a common occurrence where 18 of these agents do not matter any more.
image

We still have plans for identifying distressed agents but that is going to come out through the use of the agent config concept!

## Changes
Fixes the delete_agent graphql resolver
Performs an "upsert" of agents when registering

## Importance
From a usability perspective this is key because there is too much potential to create a large amount of agent entries thus leading to issues.

## Checklist

This PR:

### DIFF
--- a/tests/api/test_agents.py
+++ b/tests/api/test_agents.py
@@ -74,6 +74,49 @@ class TestCreateAgent:
         ).count()
         assert agent_count == 2
 
+    async def test_register_multiple_agents_label_order(self, tenant_id):
+        agent_id = await api.agents.register_agent(
+            tenant_id=tenant_id,
+            labels=["foo", "bar", "chris"],
+        )
+        agent_count = await models.Agent.where(id=agent_id).count()
+        assert agent_count == 1
+
+        agent_id = await api.agents.register_agent(
+            tenant_id=tenant_id,
+            labels=["chris", "bar", "foo"],
+        )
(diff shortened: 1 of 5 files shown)


tests/api/test_agents.py
```diff
@@ -0,0 +1,108 @@
...
+        agent = await models.Agent.where(id=agent_id).first(
+            {"id", "tenant_id", "labels", "name", "type"}
+        )
+        assert agent.id == agent_id
+        assert agent.tenant_id == tenant_id
+        assert agent.name == "MyNewAgent"
+        assert agent.type == "NewAgent"
+        assert set(agent.labels) == set(labels)
+
+    async def test_register_multiple_agents(self, tenant_id):
```
cicdw, line 50: Could you add another test similar to this one where you register an agent with the same 3 labels in different orders and confirm that only 1 new agent is actually created in the DB + the same ID is returned each time?

cicdw (changes requested):
Request for a new test, otherwise LGTM!

cicdw (approved):

PR #100, opened 2020-09-25
Changes to support the new RunConfig pattern in core

jcrist:
Adds a new `flow.run_config` field that can be used instead of `flow.environment` to configure a flow run. There's likely more to do here, but so far this works with https://github.com/PrefectHQ/prefect/pull/3333, see that PR for the full motivation.


src/prefect_server/api/flows.py
```diff
@@ -208,7 +210,8 @@ async def create_flow(
         name=flow.name,
         serialized_flow=serialized_flow,
         environment=flow.environment,
-        core_version=flow.environment.get("__version__"),
+        run_config=flow.run_config,
+        core_version=flow.__version__,
```
cicdw, line 214: I think we need `flow.__version__ or flow.environment.get("__version__")` for backwards compatibility

cicdw (commented):
I realize this won't be fully ready until we merge the Core PR, but this looks straightforward to me!  I left one comment on backwards compatibility

src/prefect_server/api/flows.py
jcrist: All flows have a `__version__` attached already (and have for a while), it's part of the base `ObjectSchema` serializer. I don't think a backwards compat fix is needed here, but might be wrong.

jcrist:
I'm seeing errors about `null` run-config values, should that field be nullable (currently it's not, but afaict the server-side default should be being applied)?

src/prefect_server/api/flows.py
cicdw: Ahh you're right

cicdw:
@jcrist the `server_default` is only used if you don't provide a value for `run_config` when inserting a new row.  But our ORM will always provide a value (and in this case it's providing `null`).  If you want to keep the non-nullability, you'll need to put a `run_config or {}` in the model before the insert.

jcrist:
I believe this is good to go.

cicdw (dismissed):
LGTM!  Welcome to the server repo @jcrist :D

cicdw (approved):

PR #103, opened 2020-10-01
Run Labels refactor

cicdw:
## Summary

This PR refactors `labels` to be specified at the flow run level (as opposed to the flow / flow group level).  Moreover, the migration will set the labels on all currently `Scheduled` runs to the correct values so that no work is missed and no backwards compatibility issues arise within the application code.

## Importance
- consolidates all our label logic into one place: `create_flow_run`
- will allow the UI to query for `flow_run{labels}` instead of recreating the flow / flow group logic as it currently does
- will allow users to configure labels on a _per run_ basis
- will allow users to schedule a flow to run multiple times simultaneously with different label sets

## Checklist

This PR:

### DIFF
--- a/services/postgres/alembic/versions/2020-09-30T123806_add_label_column_to_flow_runs.py
+++ b/services/postgres/alembic/versions/2020-09-30T123806_add_label_column_to_flow_runs.py
@@ -22,6 +22,12 @@ def upgrade():
     op.add_column(
         "flow_run", sa.Column("labels", JSONB, nullable=False, server_default="[]")
     )
+    # this query looks at all currently scheduled flow runs, or more generally any flow
+    # run that has been updated in the last day, and applies our standard label logic to it.
+    # This ensures that active flow runs don't incorrectly receive `[]` as their label set,
+    # which would cause delayed work and user frustration.
+    # Note that flow runs not touched by this query will have `[]` has their labels regardless of
+    # their flow label properties.
     op.execute(
         """
     WITH runs_to_update AS (
@@ -34,7 +40,7 @@ def upgrade():
 	from flow_run
 	join flow on flow_run.flow_id = flow.id
 	join flow_group on flow.flow_group_id = flow_group.id
-	where flow_run.state = 'Scheduled'
+	where flow_run.state = 'Scheduled' OR flow_run.updated > NOW() - interval '1 day'
 	)
     update flow_run
     set labels = runs_to_update.new_labels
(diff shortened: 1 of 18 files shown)


src/prefect_server/api/runs.py
```diff
@@ -155,6 +174,18 @@ async def _create_flow_run(
     elif flow.archived:
         raise ValueError(f"Flow {flow.id} is archived.")
 
+    # set labels
+    run_labels = []
```
jcrist, line 178: This is never hit, the `else` branch handles it.

services/postgres/alembic/versions/2020-09-30T123806_add_label_column_to_flow_runs.py
```diff
@@ -0,0 +1,48 @@
...
+down_revision = "850b76d44332"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.add_column(
+        "flow_run", sa.Column("labels", JSONB, nullable=False, server_default="[]")
+    )
+    op.execute(
```
jlowin, line 25: Just for future reference do you mind adding a small comment here - something like "for runs scheduled at the time this migration is executed, we want to make sure they have the new logic."

jlowin (changes requested):

src/prefect_server/api/runs.py
cicdw: I usually do this for readability so it's clear I expect this to be a list, but I can remove

services/postgres/alembic/versions/2020-09-30T123806_add_label_column_to_flow_runs.py
cicdw: Updated

services/postgres/alembic/versions/2020-09-30T123806_add_label_column_to_flow_runs.py
```diff
@@ -0,0 +1,54 @@
...
+
+def upgrade():
+    op.add_column(
+        "flow_run", sa.Column("labels", JSONB, nullable=False, server_default="[]")
+    )
+    # this query looks at all currently scheduled flow runs, or more generally any flow
+    # run that has been updated in the last day, and applies our standard label logic to it.
+    # This ensures that active flow runs don't incorrectly receive `[]` as their label set,
+    # which would cause delayed work and user frustration.
+    # Note that flow runs not touched by this query will have `[]` has their labels regardless of
```
jlowin, line 29:
```suggestion
    # Note that flow runs not touched by this query will have `[]` as their labels regardless of
```

jlowin (approved):

PR #106, opened 2020-10-07
Update changelog generation script to look for .yml files

joshmeek:
There are a lot of files in there that need to be cleaned up and exist with `.yml` instead of `.yaml`


jcrist (dismissed):
LGTM. Should you also run the script and update the changelog as part of this PR?

joshmeek:
@jcrist All of these old changes correspond to previously released versions of server and I'm not sure which pertains to what so I'm going to clear the directory instead of updating the changelog

cicdw (approved):
👍

PR #111, opened 2020-10-14
Allow for scheduling different parameters and run labels at the same time

cicdw:
## Summary

This PR updates the scheduler to allow for scheduling runs of the same flow at the exact same time, so long as the runs will have either different default parameter values or different run labels.  The solution relies on a somewhat messy idempotency key generation that should be stable across schedulers.

## Importance

Unlocks the last piece of the puzzle for parameter + label scheduling.

## Checklist

This PR:

### DIFF
--- a/tests/api/test_flows.py
+++ b/tests/api/test_flows.py
@@ -1142,6 +1142,10 @@ class TestScheduledRunAttributes:
             [dict(parameter_defaults=dict(x="a")), dict(labels=["b"])],
             [dict(labels=["c", "d"]), dict(labels=["c"])],
             [dict(labels=None), dict(labels=["ef"])],
+            [
+                dict(labels=None),
+                dict(labels=[]),
+            ],  # the scheduler should distinguish between none vs. empty
         ],
     )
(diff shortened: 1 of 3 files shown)


tests/api/test_flows.py
```diff
@@ -1102,6 +1102,72 @@ async def test_schedule_does_not_overwrite_flow_labels(self, project_id, labels)
...
+        flow_runs = await models.FlowRun.where({"flow_id": {"_eq": flow_id}}).get(
+            selection_set={"parameters": True, "scheduled_start_time": True},
+            order_by={"scheduled_start_time": EnumValue("asc")},
+        )
+
+        assert len(set([fr.scheduled_start_time for fr in flow_runs])) == 10
+
+    @pytest.mark.parametrize(
+        "attrs",
+        [[dict(parameter_defaults=dict(x="a")), dict(parameter_defaults=dict(x="b"))]],
```
cicdw, line 1136: Ah, I need to add some varying labels as well.  Will update to parametrize over more possibilities.

jlowin (approved):

PR #116, opened 2020-10-27
Add idempotency key to flows.create_flow

zanieb:
## Summary

Adds optional idempotency keys to the `create_flow` API, the most recent key for a flow group is stored in the `settings` dict and is referenced when creating a flow to prevent sequential flow creation with the same key.

## Importance

- Users register flows repeatedly with automation and needlessly bump versions

## Future work

This only works for **sequential** flow creation and we may want support in the future to prevent re-registrations for old flow versions. However, this would involve adding a column to the Flow table and additional complexity. We also don't really have support for old versions right now so it isn't quite necessary.

## Checklist

This PR:


tests/api/test_flows.py
```diff
@@ -261,6 +261,68 @@ async def test_flows_can_be_safely_created_twice(self, project_id, flow):
...
+
+        assert flow_id_1 == flow_id_2
+        assert flow_model_1.version == flow_model_2.version
+
+        # Verify that the flow is not duplicated
+        assert await models.Flow.where({"id": {"_eq": flow_id_1}}).count() == 1
+        # Verify that the tasks are not duplicated
+        assert await models.Task.where({"flow_id": {"_eq": flow_id_1}}).count() == len(
+            flow.tasks
+        )
```
madkinsz, line 291: Replicated from the above double creation test, may not be needed

src/prefect_server/api/flows.py
```diff
@@ -177,15 +181,22 @@ async def create_flow(
...
             ]
         }
-    ).first({"id", "schedule"})
+    ).first({"id", "schedule", "settings"})
+
+    # create the flow group or check for the idempotency key in the existing group
+    # Note: The key is stashed in the `settings` dict of the flow group which makes
+    # idempotency only apply to the most flow creation. If we want to avoid
+    # re-registering old flow versions we'll need to add a column to the Flow table
+    # TODO: Should we leave empty idempotency_keys as `None` or generate a uuid4?
```
madkinsz, line 199: Needs input

src/prefect_server/api/flows.py
```diff
@@ -194,11 +205,39 @@ async def create_flow(
...
+            flow = await models.Flow.where(
+                {
+                    "version_group_id": {"_eq": version_group_id},
+                    "archived": {"_eq": False},
+                }
+            ).first(order_by={"version": EnumValue("desc")})
+            if flow:
+                return flow.id
+            # otherwise, despite the key matching we don't have a valid flow to return
+            # and will continue as though the key did not match
```
madkinsz, line 233: I do not have coverage for this case, should I force this state to occur?

src/prefect_server/api/flows.py
joshmeek: I vote for `None` over a random UUID

src/prefect_server/api/flows.py
```diff
@@ -194,11 +205,39 @@ async def create_flow(
...
+
+        settings = flow_group.settings
+        settings["idempotency_key"] = idempotency_key
+        await models.FlowGroup.where({"id": {"_eq": flow_group.id}}).update(
+            set={"settings": settings}
+        )
```
joshmeek, line 239: This update should only be triggered if `idempotency_key` is not `None`

src/prefect_server/api/flows.py
```diff
@@ -194,11 +205,39 @@ async def create_flow(
                 "heartbeat_enabled": True,
                 "lazarus_enabled": True,
                 "version_locking_enabled": False,
+                "idempotency_key": idempotency_key,
             },
         ).insert()
     else:
         flow_group_id = flow_group.id
 
+        # check idempotency key and  early if we find a matching key and flow,
```
joshmeek, line 214: Nit: I think there's a grammatical typo in here but I can't seem to figure it out lol

src/prefect_server/api/flows.py
joshmeek: What do you mean exactly? I think a test for the case where a flow already exists using the idempotency key from the most recent creation should be added

src/prefect_server/api/flows.py
madkinsz: I'm not sure. That would allow keys to linger from past registrations so you could do something unexpected like
```
create_flow(key="foo") -> creates
create_flow(key=None) -> creates
create_flow(key="foo") -> does not create
```

joshmeek: Oh yep my bad, I understood after I wrote this that `None` is essentially an idempotency key in this sense where it only matter what the most recent create call was passing 👍

src/prefect_server/api/flows.py
madkinsz: I don't have coverage for a case where:
- The idempotency key for the flow we are creating now matches the one in the flow group settings
- All of the flows in the flow group are archived

A case exists for creating a flow with an existing idempotency key.

src/prefect_server/api/flows.py
madkinsz, line 214:
```suggestion
        # check idempotency key and exit early if we find a matching key and flow,
```

make sense this way?

joshmeek: Yep! 🙂

src/prefect_server/api/flows.py
joshmeek: Yeah I support a test case for that

src/prefect_server/api/flows.py
```diff
@@ -177,15 +181,22 @@ async def create_flow(
...
                 {"tenant_id": {"_eq": tenant_id}},
                 {"name": {"_eq": version_group_id}},
             ]
         }
-    ).first({"id", "schedule"})
+    ).first({"id", "schedule", "settings"})
+
+    # create the flow group or check for the idempotency key in the existing group
+    # Note: The key is stashed in the `settings` dict of the flow group which makes
+    # idempotency only apply to the most flow creation. If we want to avoid
```
cicdw, line 197:
```suggestion
    # idempotency only apply to the most recent flow creation. If we want to avoid
```

src/prefect_server/api/flows.py
```diff
@@ -177,15 +181,22 @@ async def create_flow(
...
+    # re-registering old flow versions we'll need to add a column to the Flow table
```
cicdw, line 198: I think we should remove this comment about old flow versions because it's much more complicated than only adding a single DB column

src/prefect_server/api/flows.py
cicdw: This could also occur if a user deletes the latest flow.

cicdw (changes requested):
A few very superficial comments

src/prefect_server/api/flows.py
madkinsz: 👍 I was thinking that note may be more appropriate in the PR anyway

src/prefect_server/api/flows.py
madkinsz: I'll add coverage for those two cases

cicdw (changes requested):
Could you add one final test?  I want to ensure we always update the idempotency key so that the following situation works as expected:
- `flow.register("Blah", idempotency_key="foo")` # v1
- `flow.register("Blah")` # v2
- `flow.register("Blah", idempotency_key="foo")` <-- should actually register a new version instead of using v2

So a test that creates a flow w/ key, creates a flow without, then creates a final flow with the original key should register a new version.

tests/api/test_flows.py
```diff
@@ -290,41 +290,60 @@ async def test_flows_not_added_with_same_idempotency_key(self, project_id, flow)
...
-    async def test_flows_added_with_different_idempotency_key(
+    @pytest.mark.parametrize(
+        "idempotency_keys",
+        [
+            pytest.param(("foo", "bar"), id="simple different keys"),
+            pytest.param((None, None), id="sequential empty keys"),
+            pytest.param(("foo", None, "foo"), id="same key with empty between"),
+            pytest.param(("foo", "bar", "foo"), id="same key with different between"),
+        ],
+    )
```
cicdw, line 301: nice, well done

cicdw (approved):
LGTM

PR #118, opened 2020-10-27
Add bulk flow run deletion mutation

### LINKED ISSUE #62: Support bulk operation for delete_flow_run mutation.
## Current behavior
The current `delete_flow_run` mutation allows one flow run to be deleted at a time. 

## Proposed behavior
Support bulk deletion of flow runs, something similar to what Hasura's auto generated schema provides. 

## Example
Please check this slack conversation: https://prefect-community.slack.com/archives/CL09KU1K7/p1598535130019500

madkinsz:
## Summary

Adds `delete_flow_runs` mutation which allows deletion of a list of flow run ids instead of calling `delete_flow_run`repeatedly

## Importance

Closes #62 

This is also a good opportunity for me to familiarize myself with the graphql routes, if this is not a feature we want to support then we can just close this ticket.

## Checklist

This PR:

### DIFF
--- a/src/prefect_server/api/runs.py
+++ b/src/prefect_server/api/runs.py
@@ -400,7 +400,7 @@ async def delete_flow_runs(flow_run_ids: List[str]) -> int:
     Deletes a set of flow runs
 
     Args:
-        - flow_run_id (str): the flow run to delete
+        - flow_run_ids (List[str]): the list of flow runs to delete
 
     Returns:
         - int: The number of affected rows
--- a/src/prefect_server/api/runs.py
+++ b/src/prefect_server/api/runs.py
@@ -394,6 +394,21 @@ async def delete_flow_run(flow_run_id: str) -> bool:
     return bool(result.affected_rows)  # type: ignore
 
 
+@register_api("runs.delete_flow_runs")
+async def delete_flow_runs(flow_run_ids: List[str]) -> int:
+    """
+    Deletes a set of flow runs
+
+    Args:
+        - flow_run_id (str): the flow run to delete
+
+    Returns:
+        - int: The number of affected rows
+    """
+    result = await models.FlowRun.where({"id": {"_in": flow_run_ids}}).delete()
+    return result.affected_rows  # type: ignore
+
+
 @register_api("runs.update_flow_run_agent")
 async def update_flow_run_agent(flow_run_id: str, agent_id: str) -> bool:
     """
--- a/src/prefect_server/graphql/runs.py
+++ b/src/prefect_server/graphql/runs.py
@@ -150,6 +150,17 @@ async def resolve_delete_flow_run(
     return {"success": await api.runs.delete_flow_run(flow_run_id=input["flow_run_id"])}
 
 
+@mutation.field("delete_flow_runs")
+async def resolve_delete_flow_runs(
+    obj: Any, info: GraphQLResolveInfo, input: dict
+) -> dict:
+    return {
+        "affected_rows": await api.runs.delete_flow_runs(
+            flow_run_ids=input["flow_run_ids"]
+        )
+    }
+
+
 @mutation.field("update_flow_run_heartbeat")
 async def resolve_update_flow_run_heartbeat(
     obj: Any, info: GraphQLResolveInfo, input: dict
--- a/src/prefect_server/graphql/schema/runs.graphql
+++ b/src/prefect_server/graphql/schema/runs.graphql
@@ -44,6 +44,9 @@ extend type Mutation {
   "Delete a flow run."
   delete_flow_run(input: delete_flow_run_input!): success_payload
 
+  "Delete a group of flow runs."
+  delete_flow_runs(input: delete_flow_runs_input!): deleted_count_payload
+
   get_runs_in_queue(input: get_runs_in_queue_input!): runs_in_queue_payload
 
 }
@@ -119,6 +122,10 @@ input delete_flow_run_input {
   flow_run_id: UUID!
 }
 
+input delete_flow_runs_input {
+  flow_run_ids: [UUID!]
+}
+
 type flow_run_id_payload {
   id: UUID
 }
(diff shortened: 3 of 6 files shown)


src/prefect_server/graphql/schema/runs.graphql
```diff
@@ -140,3 +147,7 @@ type mapped_children_payload {
   max_end_time: DateTime
   state_counts: JSON
 }
+
+type deleted_count_payload {
+  affected_rows: Int
+}
```
madkinsz, line 153: A little uncertain about best practice here

src/prefect_server/graphql/schema/runs.graphql
```diff
@@ -44,6 +44,9 @@ extend type Mutation {
   "Delete a flow run."
   delete_flow_run(input: delete_flow_run_input!): success_payload
 
+  "Delete a group of flow runs."
+  delete_flow_runs(input: delete_flow_runs_input!): deleted_count_payload
+
```
madkinsz, line 49: Generally, does it make sense to implement bulk operations by extending the original interface or adding a new route?

src/prefect_server/graphql/runs.py
```diff
@@ -150,6 +150,17 @@ async def resolve_delete_flow_run(
     return {"success": await api.runs.delete_flow_run(flow_run_id=input["flow_run_id"])}
 
 
+@mutation.field("delete_flow_runs")
+async def resolve_delete_flow_runs(
+    obj: Any, info: GraphQLResolveInfo, input: dict
+) -> dict:
+    return {
+        "affected_rows": await api.runs.delete_flow_runs(
```
madkinsz, line 158: mirroring the Hasura `affected_rows` return to inspect how many ids were deleted

tests/api/test_runs.py
```diff
@@ -786,6 +786,28 @@ async def test_delete_flow_run_fails_if_none(self, bad_value):
...
+        flow_runs = await models.FlowRun.where({"id": {"_in": flow_run_ids}}).get()
+
+        assert result == 2
+        assert list(flow_runs) == []
+
+    async def test_delete_flow_runs_allows_bad_id(self, flow_run_id):
+        result = await api.runs.delete_flow_runs(
+            flow_run_ids=[flow_run_id, str(uuid.uuid4())]
+        )
+        assert result == 1
```
madkinsz, line 805: We could error if the ids do not exist but this seems non-conformant with graphql/hasura

src/prefect_server/api/runs.py
```diff
@@ -394,6 +394,21 @@ async def delete_flow_run(flow_run_id: str) -> bool:
     return bool(result.affected_rows)  # type: ignore
 
 
+@register_api("runs.delete_flow_runs")
+async def delete_flow_runs(flow_run_ids: List[str]) -> int:
+    """
+    Deletes a set of flow runs
+
+    Args:
+        - flow_run_id (str): the flow run to delete
```
joshmeek, line 403:
```suggestion
        - flow_run_id (List[str]): the list of flow runs to delete
```

src/prefect_server/api/runs.py
```diff
@@ -394,6 +394,21 @@ async def delete_flow_run(flow_run_id: str) -> bool:
...
+
+    Returns:
+        - int: The number of affected rows
+    """
+    result = await models.FlowRun.where({"id": {"_in": flow_run_ids}}).delete()
```
joshmeek, line 408: QQ: Did you test this with a use case along the lines of where one of the IDs in the list did not exist in the db? Wondering if it still delete the other flow runs or atomically raises an error and avoids deleting any.

src/prefect_server/graphql/schema/runs.graphql
joshmeek: I'm a little more in favor of using the standard `success` payload that the other functions. Especially if the delete on a list is an atomic operation then it either deleted all of them or it didn't.

src/prefect_server/api/runs.py
madkinsz: It deletes the other flow runs / it is not atomic. I've added an explicit check in https://github.com/PrefectHQ/server/pull/118/commits/a366b49ba92214156acd29e82fdf9e52f41a79e7

src/prefect_server/api/runs.py
madkinsz, line 403:
```suggestion
        - flow_run_ids (List[str]): the list of flow runs to delete
```

src/prefect_server/graphql/schema/runs.graphql
madkinsz: It's not atomic -- does that change your mind?

src/prefect_server/api/runs.py
madkinsz: Was missing the `s` :) commited

src/prefect_server/graphql/schema/runs.graphql
joshmeek: Yeah, affected rows is fine then because it's the most convenient way to convey information on how "successful" the delete was

madkinsz: This could be moved to a generic `affected_rows` payload

cicdw (changes requested):
We don't want to support bulk operations at this time; Server users in particular have first class DB access so can perform whatever deletions they need.

There are many reasons for this, but to name a few:
- we can't support this API in Cloud and therefore shouldn't introduce it to Server either (especially considering there is a concrete workaround in the DB itself for Server users)
- we should strive for atomicity in our API, and we already regret allowing for multiple states to be set in the `set_task_run_states` API call
- error handling for bulk operations is a mess
- large changes like this need to be considered holistically before being introduced

madkinsz:
Closing per Chris' comment. Bulk operations are out of scope for server at this time.

PR #121, opened 2020-10-28
Artifacts API

joshmeek:
## Summary
Adds a new API for working with task run artifacts! This is the server side of the Artifacts API to be used in collaboration with the `artifact_api` branch on core (https://github.com/PrefectHQ/prefect/pull/3581). This adds a new table to the db `task_run_artifacts` which is a many to one relationship of artifacts to task run. Adds three new API functions and graphql mutations for creating, updating, and deleting task run artifacts.

## Checklist

This PR:

------

Server side of the Artifacts API to be used in collaboration with the `artifact_api` branch on core (https://github.com/PrefectHQ/prefect/pull/3581).

This adds a new table to the db `task_run_artifacts` which is a many to one relationship of task runs to artifacts. Adds three new API functions and graphql mutations for creating, reading, and deleting task run artifacts. Comments welcome!


services/postgres/alembic/versions/2020-10-21T141127_add_artifact_api.py
```diff
@@ -0,0 +1,53 @@
...
+
+# revision identifiers, used by Alembic.
+revision = "3c87ad7e0b71"
+down_revision = "24f10aeee83e"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.create_table(
```
cicdw, line 22: We should add an updated column + trigger to this table as well

src/prefect_server/api/artifacts.py
```diff
@@ -0,0 +1,80 @@
...
+        - task_run_id (str): the task run id
+        - kind (str): the artifact kind
+        - data (dict): the artifact data
+        - tenant_id (str, optional): the tenant id that this artifact belongs to. Defaults
+            to the tenant ID linked to the task run
+
+    Returns:
+        - str: the task run artifact ID
+    """
+    task_run = await models.TaskRun.where(id=task_run_id).first({"tenant_id"})
```
cicdw, line 25: It's important to check whether `task_run_id` is `None` before running this ORM call - Hasura treats `None` as "no where clause at all" so this would grab a random run and _possibly_ cause performance issues.

And just because it avoids a DB scan, I usually do:
```python
if not task_run_id: # handles empty strings as well, why not short-circuit
```

src/prefect_server/api/artifacts.py
```diff
@@ -0,0 +1,80 @@
...
+
+    if not task_run:
+        raise ValueError(f"Creating artifact failed for task run ID {task_run_id}")
+
+    if not tenant_id:
```
cicdw, line 30: It's good practice to also check whether the provided tenant id matches the one on the task run, and raise an error if they don't match

src/prefect_server/api/artifacts.py
```diff
@@ -0,0 +1,80 @@
...
+        - data (dict): the artifact data
+
+    Returns:
+        - bool: if the update was successful
+    """
+    if not task_run_artifact_id:
+        raise ValueError("Must supply a valid task run artifac ID to update")
+
+    result = await models.TaskRunArtifact.where(id=task_run_artifact_id).update(
+        set={"data": data}
```
cicdw, line 58: I don't necessarily have pushback, but curious why you chose a full update API instead of an append API?

src/prefect_server/api/artifacts.py
```diff
@@ -0,0 +1,80 @@
...
+    Delete a task run artifact.
+
+    Args:
+        - task_run_artifact_id (str): the task run artifact id
+
+    Returns:
+        - bool: if the deletion was successful
+    """
+    if not task_run_artifact_id:
+        raise ValueError("Must supply a valid task run artifac ID to delete")
```
cicdw, line 77:
```suggestion
        raise ValueError("Must supply a valid task run artifact ID to delete")
```

src/prefect_server/graphql/artifacts.py
```diff
@@ -0,0 +1,50 @@
...
+@mutation.field("create_task_run_artifact")
+async def resolve_create_task_run_artifact(
+    obj: Any, info: GraphQLResolveInfo, input: dict
+) -> dict:
+
+    data = input.get("data")
+
+    data_size = sys.getsizeof(json.dumps(data))
+    if data_size > 1000000:  # 1 mb max
+        raise ValueError("Artifact data payload exceedes 1Mb limit.")
```
cicdw, line 21: nice

src/prefect_server/api/__init__.py
```diff
@@ -8,3 +8,5 @@
 import prefect_server.api.runs
 import prefect_server.api.messages
 import prefect_server.api.agents
+import prefect_server.api.artifacts
+import prefect_server.api.deprecated
```
cicdw, line 12: what's this `prefect_server.api.deprecated` for?

cicdw (changes requested):
Nice!  I left some small comments / questions / requests

src/prefect_server/api/artifacts.py
madkinsz: I'm curious what you mean by an append API

services/postgres/alembic/versions/2020-10-21T141127_add_artifact_api.py
joshmeek: When I spoke to @jlowin about that column he advised against adding it

src/prefect_server/api/__init__.py
joshmeek: Oh I think that's a merge conflict accidental add, will remove

services/postgres/alembic/versions/2020-10-21T141127_add_artifact_api.py
jlowin: I think I may have inadvertently created confusion here -- we do need this column/trigger on every table; for the most highly-trafficked tables (which this will certainly not be), that can create heavy update contention which has to be dealt with separately.

joshmeek: Okay awesome, thanks for clarifying 👍  will add this column/trigger

src/prefect_server/api/artifacts.py
cicdw: Right now if a task run wants to publish two artifacts, it'll need to collect them _both_ client-side and call `update_task_run_artifact` with the full payload.  

Alternatively, we _could_ change the API behavior so that calling `update_task_run_artifact` allows users to add multiple artifacts in sequence (this might require an implicit schema-change to the JSONB in the `data` column too).

madkinsz: Perhaps I'm misunderstanding what was done but the mapping for task_run -> artifacts is one -> many so they actually call `create_task_run_artifact` each time they want to add/append an artifact and `update_task_run-artifact` exists solely so an already created task run artifact can have its data replaced (for which an example use-case is a progress bar).

joshmeek: I don't think we should support append behavior for the update. If a task is going to publish two artifacts then it is going to have to call `create` twice anyways in order to create the two artifacts. Client side enforcement of how the data should be updated is probably better than reading the data and determining what to append. 

A downside of append behavior I could think of is something like this: if I have a piece of data that is of `markdown` type and I want to update where do I append the data? Is it a new entry in the data blob or do I merge the two strings together? Now pieces like the UI have to decide how to render artifacts with multiple markdown values. (which one gets displayed first, etc.) I could see the same thing with link artifacts where ideally it should be one link artifact = one link, overwrite only and we don't have to have any sense of resolution when it comes to merging the data.

cicdw: Ahhh yes thank you both for clarifying; I am 100% on the same page now and definitely agree your current API is the way to go 👍

src/prefect_server/api/artifacts.py
```diff
@@ -0,0 +1,86 @@
...
+    if not task_run_id:
+        raise ValueError("A `task_run_id` must be provided to create artifact")
+
+    task_run = await models.TaskRun.where(id=task_run_id).first({"tenant_id"})
+
+    if not task_run:
+        raise ValueError(f"Task run {task_run_id} not found")
+
+    if tenant_id and tenant_id != task_run.tenant_id:
+        raise ValueError("Provided tenant ID does not match task run")
```
cicdw, line 34:
```suggestion
        raise ValueError("Invalid tenant ID.")
```
We don't want to give the impression that you can validate a given tenant ID by spamming this route so instead simply saying "Invalid" is better

cicdw (changes requested):
One minor request for a different error message, otherwise this LGTM!

cicdw (approved):
LGTM - nice work

PR #57, opened 2020-08-25
WIP: Helm chart

shaunc:
## Summary

This helm chart is a work in progress (not for merging -- except into separate branch if appropriate). Will create database and init with schema. All services come up except Apollo.

**If you have some time, please give me some help with apollo.** Currently failing with:
```
kc logs --follow prefect-server-apollo-bb688cf9f-5fc9j 
{"status":"ok","version":"UNKNOWN"}
GraphQL service healthy!

> @ serve /apollo
> node dist/index.js

Building schema...
Error: Could not obtain introspection result, received: {"path":"$","error":"resource does not exist","code":"not-found"}
    at getSchemaFromIntrospection (/apollo/node_modules/@graphql-tools/wrap/index.cjs.js:1370:15)
    at introspectSchema (/apollo/node_modules/@graphql-tools/wrap/index.cjs.js:1379:12)
    at process._tickCallback (internal/process/next_tick.js:68:7)
Trying again in 3 seconds...
Building schema...
Error: Could not obtain introspection result, received: {"path":"$","error":"resource does not exist","code":"not-found"}
    at getSchemaFromIntrospection (/apollo/node_modules/@graphql-tools/wrap/index.cjs.js:1370:15)
    at introspectSchema (/apollo/node_modules/@graphql-tools/wrap/index.cjs.js:1379:12)
    at process._tickCallback (internal/process/next_tick.js:68:7)
Trying again in 3 seconds...
```

## Changes

Adds experimental helm chart.

## Importance

Allows deployment of prefect server to kubernetes.

## Checklist

This PR:

- [ ] adds new tests (if appropriate)
- [ ] adds a change file in the `changes/` directory (if appropriate)

PR #123, opened 2020-11-02
Add a Helm chart for deployment

madkinsz:
## Summary

Adds a Helm chart to deploy Prefect Server to Kubernetes.

Branches off of #57. Thank you to the community for getting it started, especially @shaunc! This notably differs in the removal of deployment of ingresses which require an ingress controller to work and is not something we intend to provide out of the box right now.

See the rendered README at https://github.com/PrefectHQ/server/blob/pr/57/helm/prefect-server/README.md

## Importance

- Improve accessibility to deployment on clusters
- Demonstrate best practices for deployments

## Checklist

This PR:

## Future work
- Release to a helm repo with pinned `prefectVersionTag`
- Improve documentation
- Respond to user feedback
- Add automated testing i.e. calling helm template then checking for expected values


madkinsz:
Example NOTES:

```
- Run the following command to get the UI URL:

  UI_HOST=$( \
    kubectl get svc \
    --namespace default \
    --template "{{ range (index .status.loadBalancer.ingress 0) }}{{.}}{{ end }}" \
    my-release-ui \
  ) \
  && echo "UI available at: http://$UI_HOST:8080"

  NOTE: It may take a few minutes for the LoadBalancer IP to be available. 
        You can watch the status of by running: kubectl get --namespace default svc -w my-release-ui

- Run the following command to get the Apollo GraphQL API URL:

  API_HOST=$( \
      kubectl get svc \
      --namespace default \
      --template "{{ range (index .status.loadBalancer.ingress 0) }}{{.}}{{ end }}" \
      my-release-apollo \
    ) \
    && echo "API available at: http://$API_HOST:4200/graphql"

- The UI has been configured to point to 'http://localhost:4200/graphql' by default. 
  - The API location you retrieved above should match this url. 
  - The default can be changed in the helm deployment at 'ui.apolloApiUrl' if it is incorrect. 
  - The location can also be changed within the UI itself.
  - The API must be accessible from the the user's machine for the UI to work
```

madkinsz:
I could create something like https://github.com/PrefectHQ/server/pull/123/commits/ebec1630fb7ffc7a9509a675d7c2712d96465b48 for many of the things that need component/global merge. I could also abstract huge chunks of the deployments/services since they're all about the same but not sure how far it's worth going within helm.

helm/prefect-server/templates/_helpers.tpl
```diff
@@ -0,0 +1,262 @@
+{{/*
```
jcrist, line 1: There's a lot of magic going on in this file (most of these appear to be borrowed from JupyterHub). While fine, I find the patterns here hard to read and understand what's going on - would prefer a more explicit approach when possible.

To be concrete, take `prefect-server.matchLabels`. The generated values for this template helper change dependent on the file they're used in (and the scope in which they're used), which is different than how most template helpers work (usually they're just shorthand to avoid repeating the same string-generating code everywhere). I had to dig deep to determine that the `matchLabels` used in the apollo deployment/service are different than those in the graphql deployment/service.

I think I'd prefer if the templates here were more explicit. It will mean that service-specific things (like adding the component name to the match labels) will need to be done in each service, but it makes the templates easier to read IMO.

helm/prefect-server/values.yaml
```diff
@@ -0,0 +1,273 @@
...
+
+  service:
+    # type defines the service type and defaults to ClusterIP
+    # because this service does not need to be exposed outside
+    # the cluster
+    type: ClusterIP
+
+  labels: {}
+  annotations: []
+  replicaCount: 1
```
jcrist, line 138: Suggest to use `replicas` instead of `replicaCount` everywhere, since that's what k8s uses.

helm/prefect-server/values.yaml
```diff
@@ -0,0 +1,273 @@
...
+    create_pgcrypto.sql: |
+      -- create pgcrypto extension, required for Hasura UUID
+      CREATE EXTENSION IF NOT EXISTS pgcrypto;
+      CREATE EXTENSION IF NOT EXISTS "pg_trgm";
+      SET TIME ZONE 'UTC';
+
+# prefectConfig allows override of prefect_server/config.toml values
+# Values placed here will result in environment variable definitions
+# starting with "PREFECT_SERVER__" and will be passed to TODO
+prefectConfig:
```
jcrist, line 95: Instead of hardcoding some values, I'd suggest we leave this blank and put anything that is required into the images (I think this is already done). If this is more to document what a user might want to change, we can always handle that in docs or in a comment.

The issue with hardcoding values here is that if we change them in the code (update the `config.toml` in prefect-server) we'd also need to change them here to have them take effect.

helm/prefect-server/values.yaml
```diff
@@ -0,0 +1,273 @@
...
+# which creates a graphql api from the postgres database
+hasura:
+  port: 3000
+
+  # hasura.image configures the docker image used for hasura
+  # and is the only image in the chart that is not hosted by
+  # Prefect
+  image:
+    repository: hasura/graphql-engine
+    version: v1.3.0
```
jcrist, line 123:
```suggestion
    name: hasura/graphql-engine
    tag: v1.3.0
```

`name` and `tag` are more common here (and `tag` is more accurate than `version`.

helm/prefect-server/values.yaml
```diff
@@ -0,0 +1,273 @@
...
+    resources: {}
+
+# apollo configures the Prefect apollo deployment and service
+# which provides a unified graphql schema for users and the UI
+# to interact with
+apollo:
+  port: 4200
+  image:
+    repository: prefecthq/apollo
+    version: null
```
jcrist, line 184:
```suggestion
    name: prefecthq/apollo
    tag: null
```

helm/prefect-server/values.yaml
```diff
@@ -0,0 +1,273 @@
...
+  nodeSelector: {}
+  tolerations: []
+  affinity: {}
+
+# ui configures the Prefect ui deployment and service
+ui:
+  port: 8080
+  image:
+    repository: prefecthq/ui
+    version: null
```
jcrist, line 213:
```suggestion
    name: prefecthq/ui
    tag: null
```

helm/prefect-server/values.yaml
```diff
@@ -0,0 +1,273 @@
...
+  nodeSelector: {}
+  tolerations: []
+  affinity: {}
+
+# towel configures the Prefect towel deployment which provides
+# a group of useful services
+towel:
+  image:
+    repository: prefecthq/server
+    version: null
```
jcrist, line 246:
```suggestion
    name: prefecthq/server
    tag: null
```

helm/prefect-server/values.yaml
```diff
@@ -0,0 +1,273 @@
+# prefect-server configuration
+
+# image provides configuration for all prefect images
+# this notably excludes the hasura image
+image:
+  version: "latest"
+  pullPolicy: IfNotPresent
```
jcrist, line 7: Why add shared defaults for all prefect images? IMO this adds extra complications that we may not need. A user is likely to upgrade the helm chart (which will bump the release), or want to explicitly set things for specific deployments (say pin the UI to a version), I don't see the need for a shared setting like this.

That said, if we keep this we should change the `image` field to something more descriptive like `imageDefaults` or something. But I think we should drop this for now - can always add it back later if people ask for it.

helm/prefect-server/values.yaml
```diff
@@ -0,0 +1,273 @@
...
+  tolerations: []
+  affinity: {}
+  init:
+    resources: {}
+
+# apollo configures the Prefect apollo deployment and service
+# which provides a unified graphql schema for users and the UI
+# to interact with
+apollo:
+  port: 4200
```
jcrist, line 181: Typically helm charts will use fixed internal ports (e.g. the apollo pods would all listen on a hardcoded port inside their pod), and only external-facing ports are configurable. I didn't trace through the helm chart deeply, but on first glance it looks like you're exposing both (the port here for setting the pod-internal ports, and the one in the service for external-facing ports).

I suggest we only expose the external `service` ports, and hardcode the internal ones - the user should have no need to configure these.

helm/prefect-server/templates/_helpers.tpl
madkinsz: re JupyterHub, I copied their styling for docs and used parts of their `componentName` / `nameField` functions but I tried to stay away from duplicating the rest of what they have because I agree it's complicated and they're triple licensed and I didn't want to deal with that.

For the specific example, `prefect-server.matchLabels` could be less magical in a few ways (of increasing verbosity)
1. Extend the docstring to explain that the component name is used, this should be done no matter what
2. Add the componentName to the scope at the top of each template (I do not know how to do this, seems nearly as magical as file name inference)
3. Pass the component name explicitly e.g. `{{- include "prefect-server.matchLabels" (merge (dict "componentName" "hasura") .) -}}`
4. Write a helper for each component that repeats the exact same merge logic but with a hard-coded componentName

4 is what the original PR did and I am pretty against it because then the logic has to be repeatedly changed
3 is okay but the `merge` syntax is pretty verbose and has to be repeated for every item in the deployment/object that contains the component name. I could be convinced though.

I'd argue that adding more documentation to the top of _helpers explaining that we infer component names and doing so consistently is quite clear. I think Helm charts are generally hard to read and I want to make this as clear as possible but I'm not sure any user is actually going to look at how labels are merged, they'll just see that they can define global and component specific labels in the config.

helm/prefect-server/values.yaml
madkinsz: I'll try removing these, they were in the original PR and I agree they shouldn't be hardcoded if they match the current config.

helm/prefect-server/values.yaml
madkinsz: So, I'm presuming that we will not be using "latest" here but pin to a version on release. If we do want to use "latest", the following argument doesn't really apply and I'm happy to move them all into service specific config.

The shared defaults exist because we generally release all of the images together and I don't think that it makes sense to update ~5 image tags each time we bump the helm version. If the user wants to pin the UI that's still straight-forward and I think it's easier for them to say:
- I want to pin these two versions so I will change them within the per-service configs
- The rest I want to leave so those values will remain empty

Otherwise when they pin a couple versions it'll get confusing to merge with updates from us.

helm/prefect-server/templates/_helpers.tpl
jcrist: Another option (and what I'd recommend) is to have a common set of `matchLabels` used for all deployments, and only add the component name at the end explicitly. Simple, easy to read (I too find the `merge` + `dict` pattern complicated), and doesn't require looking into the template file to understand. This is what I do in dask-gateway and have seen in other charts:

https://github.com/dask/dask-gateway/blob/c2583548df19359d24031e1dd9161c616d3bed50/resources/helm/dask-gateway/templates/gateway/deployment.yaml#L10-L17

madkinsz: That is cleaner, I like that. However the component name is being used twice and is composed in the first one e.g.
```
name: prefect-server-apollo
component: apollo
```

https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/#labels is not particularly clear here. Should the `name` be just `apollo`? Should the `component` be `api`?

helm/prefect-server/values.yaml
jcrist: > The shared defaults exist because we generally release all of the images together and I don't think that it makes sense to update ~5 image tags each time we bump the helm version.

This could be automated, and is also easier to understand (IMO) than a hierarchy of version settings. I'd rather optimize for user usability than shortness of diffs for us.

> Otherwise when they pin a couple versions it'll get confusing to merge with updates from us.

I don't follow. Without a global setting, they'd pin the UI image tag, and leave the others blank (which would default to the versions shipped with the helm chart). When they want to upgrade, they'd upgrade the helm chart and the unpinned images would upgrade as intended.

I don't think it's a good idea to encourage users to upgrade the prefect server without also bumping the chart - some new server versions may require a change to the chart (new config, new secrets, etc...). Exposing the image name and tag is mostly useful IMO for users who want to build their own images, we should still release a new chart with every prefect-server release IMO.

---

That said, if you want to go with your version I'd suggest:

- Remove the global `pullPolicy`, let users set that per-component only
- Remove the nesting and rename as `prefectVersion`, with a comment that this will be used for the `tag` for all prefect images unless overwritten per component.

I found the `image` grouping confusion, since it wasn't fully specifying an image, only defaults for each sub-components. Simplifying to the above should satisfy your goals but is less confusing to me.

helm/prefect-server/templates/_helpers.tpl
madkinsz: Also note that this is a simple case. What about something like 'annotations' where we want to merge global and local config values?

jcrist: Things like `app.kubernetes.io/name` and `app.kubernetes.io/instance` should refer to the whole application (`prefect-server`), while `app.kubernetes.io/component` should be things like `apollo` or `graphql`. The important thing here is that pods in a deployment have a unique set of `matchLabels` to identify them as different from other deployments.

jcrist: > What about something like 'annotations' where we want to merge global and local config values?

I'd probably drop the global and local config thing, and go for only local for most of these. I can't think of many good use cases for wanting to apply a blanket set of things across all objects - most things a user will want to tweak will be local to a specific component (e.g. add an annotation to pods in the apollo deployment).

That said, I guess what I'm arguing for here is that repeating yourself in each deployment seems clearer to me than doing some magical inferences inside a helper template. There's some readability cost to indirection here that I'm trying to avoid.

shaunc: Thank you everyone for moving this forward! I got it to the state of "works for me" and promptly got absorbed in a machine learning project -- which happily uses the current version. I think @jcrist is on the right track. When I wrote the helpers my concern wasn't so much transparency for the user as that I knew prefect server itself was rapidly evolving -- and I didn't understand the details. The "abstract" helpers and global/local split were meant to allow prefect server itself to change without tracking down duplicated settings. You are all closer to prefect-server itself, and don't have to be so "arms length".

One exception is that -- if you happen to want to run multiple copies of prefect server on one cluster (e.g. for different clients, and/or using different ingresses/load balancers) there are some things that you will want to inject globally -- I think of annotations, in particular. (Maybe affinities/tolerations as well.)

madkinsz: > Things like `app.kubernetes.io/name` and `app.kubernetes.io/instance` should refer to the whole application (`prefect-server`), while `app.kubernetes.io/component` should be things like `apollo` or `graphql`.

This is in disagreement with the above K8s docs in which `part-of` refers to the whole application and `name` refers to the name of the part.

> I'd probably drop the global and local config thing, and go for only local for most of these. I can't think of many good use cases for wanting to apply a blanket set of things across all objects - most things a user will want to tweak will be local to a specific component (e.g. add an annotation to pods in the apollo deployment).

I'm not a Helm user but I presumed adding annotations or pull secrets for the whole deployment of our server would be common. I can drop it if that's not the case.

> That said, I guess what I'm arguing for here is that repeating yourself in each deployment seems clearer to me than doing some magical inferences inside a helper template. There's some readability cost to indirection here that I'm trying to avoid.

I think it's worth pointing out that the helper template is only ~200 lines and this deployment is pretty simple. I think repeated logic is harder to maintain, I made these changes because working with it in the original PR was a pain.

I'm not sure what to do here. I think the best solution is to see what it actually looks like / what is needed once it's in use and I'm hesitant to do a bunch more work to simplify it when I may have to add it back in later.

helm/prefect-server/values.yaml
madkinsz: Yeah I agree the `image` grouping is poorly named. I am a bit confused by this whole bit and presume it's because I've not done any helm upgrading of charts -- I'll just move to per service image settings and leave them at 'latest'  by default and we can write automation to pin them to a version in the future.

changes/pr123.yaml
```diff
@@ -0,0 +1,2 @@
+feature:
+  - "Add experimental Helm chart for deploying in K8s - [#123](https://github.com/PrefectHQ/server/pull/123)"
```
cicdw, line 2: Would you mind including a `contributor` section here that lists @shaunc , @yolibernal and @joelluijmes as contributors?  Definitely want to recognize their effort in kicking this off and QA'ing

cicdw (changes requested):
Minor request for changelog update; also noticed `helm/prefect-server/charts/postgresql-9.3.2.tgz` was included - was that intentional?

I'll give another pass before merging but overall this LGTM

changes/pr123.yaml
madkinsz: Great idea! Will do

madkinsz: > Minor request for changelog update; also noticed `helm/prefect-server/charts/postgresql-9.3.2.tgz` was included - was that intentional?

I tried removing it and then `helm install ...` failed, I'll look into it a bit more.

madkinsz: Fixed :)

jcrist: Good catch, I didn't notice that. Yeah, we shouldn't vendor this, dependencies should be managed externally. Note that just deleting the file won't be sufficient to reduce the repo size (not sure if we care about this), since the commit that adds it will still be present.

madkinsz: I did a rebase and dropped it from the commit that added it -- my git-foo is only medium strength though so perhaps worth someone else checking it's gone.

cicdw: It's still in the history, but it's also only 44k in size so I think it's OK

cicdw (approved):
Awesome - nice work!

tonycpsu:
@madkinsz : Is there a reason ingress support was removed?  Running two application load balancers in cloud environments can be cost-prohibitive for customers with smaller deployments, and creates a potential security issue unless steps are taken to lock down the endpoints.  Is the configuration with load balancers for both the UI and Apollo servers the only one supported?  If so, I think that will limit adoption from people who are test-driving Prefect as compared to Airflow and other options.

madkinsz:
Hi @tonycpsu, to support ingresses we must make assumptions about the presence of an ingress controller and it seemed out of scope for this chart to manage an nginx subchart. It also significantly complicated the chart since we had to manage ingress hosts. We're not really helm chart experts, so I'm not sure what standard practice here is. If you can point me to a deployment with several components like this that supports ingresses well, I can take a look at the feasibility.

tonyc-blastpoint:
I'm not a Helm chart expert either, but good artists copy, great artists steal, right?  Airflow's [chart](https://github.com/apache/airflow/blob/master/chart/) seems to [support](https://github.com/apache/airflow/blob/master/chart/values.yaml#L46) ingresses.

The thing I ran into when trying to add ingresses on my own (and tried to ask about on the Prefect Slack but did not get a response to) is that the Prefect UI server doesn't seem to like having its URLs rewritten, and rewriting is kind of essential to how ingresses work, since it involves multiple services (in this case, at least Prefect and GraphQL) sharing the same HTTP/HTTPS server using different URL prefixes.

In the mean time, since ingresses aren't supported...  Is there a way to run using a NodePort configuration instead of ClusterIP or LoadBalancer, so that we can have internal access to the Prefect UI without having to port forward or make use of expensive AWS load balancers?

tonycpsu:
I'm not a Helm chart expert either, but good artists copy, great artists steal, right?  Airflow's [chart](https://github.com/apache/airflow/blob/master/chart/) seems to [support](https://github.com/apache/airflow/blob/master/chart/values.yaml#L46) ingresses.

The thing I ran into when trying to add ingresses on my own (and tried to ask about on the Prefect Slack but did not get a response to) is that the Prefect UI server doesn't seem to like having its URLs rewritten, and rewriting is kind of essential to how ingresses work, since it involves multiple services (in this case, at least Prefect and GraphQL) sharing the same HTTP/HTTPS server using different URL prefixes.

In the mean time, since ingresses aren't supported...  Is there a way to run using a NodePort configuration instead of ClusterIP or LoadBalancer, so that we can have internal access to the Prefect UI without having to port forward or make use of expensive AWS load balancers?

madkinsz:
> I'm not a Helm chart expert either, but good artists copy, great artists steal, right? Airflow's [chart](https://github.com/apache/airflow/blob/master/chart/) seems to [support](https://github.com/apache/airflow/blob/master/chart/values.yaml#L46) ingresses.
> 
> The thing I ran into when trying to add ingresses on my own (and tried to ask about on the Prefect Slack but did not get a response to) is that the Prefect UI server doesn't seem to like having its URLs rewritten, and rewriting is kind of essential to how ingresses work, since it involves multiple services (in this case, at least Prefect and GraphQL) sharing the same HTTP/HTTPS server using different URL prefixes.

This makes it sound like implementing an ingress is non-trivial and I'm not sure if we have the bandwidth to maintain something like ingresses on a kubernetes deployment of Server which is already a pretty advanced use-case. I'll definitely keep this in mind though, if I have spare time I can look into it further. Note to anyone else, airflow added ingress support in https://github.com/apache/airflow/pull/10064

> In the mean time, since ingresses aren't supported... Is there a way to run using a NodePort configuration instead of ClusterIP or LoadBalancer, so that we can have internal access to the Prefect UI without having to port forward or make use of expensive AWS load balancers?

I believe you can switch to a NodePort without much trouble but I'll have to test it out if you want a guide on how to do so. I've opened an issue so I can track this.

shaunc:
When I originally contributed the PR that started this off, I copied ingress support guided by (a subset) of [ingress support options in devspace](https://devspace.sh/component-chart/docs/configuration/ingress). It wasn't much work, and worked/works well for me. nginx-ingress, at least, does almost everything with annotations, so the chart itself doesn't need much logic to support.

jcrist:
Yeah, I think having an ingress shouldn't be on by default (or required), but supporting an opt-in option which requires some user configuration makes sense to me. This is what JupyterHub does as well, and it seems to work well. https://zero-to-jupyterhub.readthedocs.io/en/latest/administrator/advanced.html#ingress, https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/master/jupyterhub/templates/ingress.yaml.

As noted above, this _may_ require some changes to the UI to support running behind a prefix if that's not already working (not sure, cc @znicholasbrown ).

Issue #120, opened 2020-10-28
Expose Prefect Core Version on Server/UI


AlessandroLollo:
## Current behavior

Actually, there's no easy way to get the version of Prefect Core running in my environment.

## Proposed behavior

Expose Prefect Core version so that it can be shown in Prefect UI/Server.

## Example

Having Prefect Core version under the Prefect Core logo would be nice!
![Screenshot 2020-10-28 at 14 31 40](https://user-images.githubusercontent.com/1206243/97442826-8f64c980-192a-11eb-8dce-2d4af1858010.png)

madkinsz:
@znicholasbrown should this be in the ui repo?

znicholasbrown:
Hm no @madkinsz, I think the version will need to be added from Core and passed as an environment variable to the UI or Server containers, the latter is preferable since the UI can run independently of Server. If we go that way, we'd want to expose the version through the `api` route... we already do this but it's represented as a commit hash instead of a semver.

madkinsz:
> Hm no @madkinsz, I think the version will need to be added from Core and passed as an environment variable to the UI or Server containers, the latter is preferable since the UI can run independently of Server. If we go that way, we'd want to expose the version through the `api` route... we already do this but it's represented as a commit hash instead of a semver.

Great, thanks for the background. I'll look into this a bit more.

madkinsz:
@AlessandroLollo heads up that this done!

AlessandroLollo:
Great!

AlessandroLollo:
> @AlessandroLollo heads up that this done!

Great!

PR #126, opened 2020-11-09
Add `core_version` to server api query

madkinsz:
## Summary

Adds a human readable version string to the server api query indicating the Prefect Core version.

## Importance

Displaying the core version the Server is using in the UI is helpful.

Gets going on #120 -- Will need UI update

## Checklist

This PR:


cicdw (commented):
Code wise this looks good; before we move forward with this implementation though, I want to raise a UX issue I foresee: I suspect most users use `prefect server start` to stand up Server.  The images that this uses are always built prior to tagging `prefect` for release, meaning that if I'm on 0.13.4 of Prefect (for example), I'll see something like `0.13.3+63.g4c764ddb8` as my `core_version` which I suspect will confuse people.

I'm not 100% sure how to avoid this chicken / egg problem

znicholasbrown:
Could we add an env variable to the build step (something like `target_release`) to use for official releases but fall back to the commit hash for non-tagged spinups?

madkinsz:
> Code wise this looks good; before we move forward with this implementation though, I want to raise a UX issue I foresee: I suspect most users use `prefect server start` to stand up Server. The images that this uses are always built prior to tagging `prefect` for release, meaning that if I'm on 0.13.4 of Prefect (for example), I'll see something like `0.13.3+63.g4c764ddb8` as my `core_version` which I suspect will confuse people.
> 
> I'm not 100% sure how to avoid this chicken / egg problem

I see two options:
- Rebuild the docker images with the correct version of Prefect Core, it seems a little weird that we are retagging when the core code could be out of date anyway
- Pass an environment variable in the `prefect server start` docker-compose and use that if present e.g.

```python
        "version": os.getenv("PREFECT_SERVER_VERSION", prefect_server.__version__),
        "core_version": os.getenv("PREFECT_CORE_VERSION", prefect.__version__),
```
> Could we add an env variable to the build step (something like `target_release`) to use for official releases but fall back to the commit hash for non-tagged spinups?

The build step for the docker image? This would still be out of date like Chris mentioned.

cicdw:
I vote we take the environment variable route: 
```python
"core_version": os.getenv("PREFECT_CORE_VERSION", prefect.__version__),
```

Purely as FYI: the only thing that Server relies on from Prefect Core is serialization logic for state objects and some Flow objects, so more often than not we don't need to touch the Server images when we make core changes.  Given that, I think this `core_version` field is more of a "schema compatibility" signal, so you know what version of Core "sanctioned" this particular server image.

changes/pr126.yaml
```diff
@@ -0,0 +1,2 @@
+feature:
```
cicdw, line 1:
```suggestion
enhancement:
```
as a rule of thumb we usually reserve `feature` for the big big stuff (e.g., the helm chart) and `enhancement` for the rest

cicdw (changes requested):
really minor request to re-categorize as `enhancement,` otherwise this LGTM!

madkinsz:
Note this will require a PR in `prefect` to pass this variable and a PR in `ui` to display it.

cicdw (approved):

PR #130, opened 2020-11-11
Fix dev infrastructure spinup failure

zanieb:
## Summary

`prefect-server dev infrastructure` fails because the database isn't ready, the check on the exception waiting for readiness was made too narrow.

## Changes

- Adds an additional case to the exception catch that waits for the database to be ready

Was failing with:
```
Database upgrade encountered fatal error:

(psycopg2.OperationalError) server closed the connection unexpectedly
	This probably means the server terminated abnormally
	before or while processing the request.

(Background on this error at: http://sqlalche.me/e/13/e3q8)

Exception caught; killing services (press ctrl-C to force)
```

## Importance

- Spinning up dev infrastructure was broken

## Checklist

This PR:

### DIFF
--- a/changes/pr130.yaml
+++ b/changes/pr130.yaml
@@ -1,0 +1,2 @@
+fix:
+  - "prefect-server dev infrastructure would sometimes fail to start with 'connection closed' - [#130](https://github.com/PrefectHQ/server/pull/130)"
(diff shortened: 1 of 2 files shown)


cicdw (approved):
good catch

PR #134, opened 2020-11-12
Fix scheduling duplicate runs with expired idempotency keys

joshmeek:
## Summary
Through https://github.com/PrefectHQ/prefect/issues/3629 we observed that there is potential for duplicate runs to be scheduled due to expired idempotency keys (24 hours). To patch this behavior this PR moves the last scheduled run check to the step prior to creating an idempotency key based on the parameter defaults and labels.

## Importance
Duplicate runs are a hugely unexpected behavior and this patch is necessary.

## Checklist

This PR:

### DIFF
--- a/src/prefect_server/api/flows.py
+++ b/src/prefect_server/api/flows.py
@@ -579,21 +579,6 @@ async def schedule_flow_runs(flow_id: str, max_runs: int = None) -> List[str]:
     # schedule every event with an idempotent flow run
     for event in flow_schedule.next(n=max_runs, return_events=True):
 
-        # TODO: This has potential to miss scheduling runs that all need to be scheduled at the same time
-        # in the case of a restart while only a portion of that time's runs have been scheduled. There
-        # needs to be some external tracker or query here that looks to see if any of the runs have
-        # already been scheduled.
-        #
-        # For example, say there is a flow with a schedule that has 10 different clocks, each with a
-        # different parameter. This loop is running through to schedule the next 10 runs and half way
-        # through the service restartes (due to some external pressure). Then when it loads back up
-        # it will not schedule the remaining half of the runs because of the last_scheduled_run check
-        # below. Therefore something needs to check if all of the runs for that particular time had
-        # actually been scheduled.
-
-        # if this run was already scheduled, continue
-        if last_scheduled_run and event.start_time <= last_scheduled_run:
-            continue
         # if the event has parameter defaults or labels, we do allow for
         # same-time scheduling
         if event.parameter_defaults or event.labels is not None:
@@ -604,6 +589,9 @@ async def schedule_flow_runs(flow_id: str, max_runs: int = None) -> List[str]:
             idempotency_key = (
                 f"auto-scheduled:{event.start_time.in_tz('UTC')}:{md5.hexdigest()}"
             )
+        # if this run was already scheduled, continue
+        elif last_scheduled_run and event.start_time <= last_scheduled_run:
+            continue
         else:
             idempotency_key = f"auto-scheduled:{event.start_time.in_tz('UTC')}"
 
--- a/src/prefect_server/api/flows.py
+++ b/src/prefect_server/api/flows.py
@@ -579,6 +579,18 @@ async def schedule_flow_runs(flow_id: str, max_runs: int = None) -> List[str]:
     # schedule every event with an idempotent flow run
     for event in flow_schedule.next(n=max_runs, return_events=True):
 
+        # TODO: This has potential to miss scheduling runs that all need to be scheduled at the same time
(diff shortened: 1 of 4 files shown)


src/prefect_server/api/flows.py
```diff
@@ -579,19 +579,31 @@ async def schedule_flow_runs(flow_id: str, max_runs: int = None) -> List[str]:
...
+        # below. Therefore something needs to check if all of the runs for that particular time had
+        # actually been scheduled.
+
+        # if this run was already scheduled, continue
+        if last_scheduled_run and event.start_time <= last_scheduled_run:
+            continue
         # if the event has parameter defaults or labels, we do allow for
         # same-time scheduling
-        if event.parameter_defaults or event.labels is not None:
+        elif event.parameter_defaults or event.labels is not None:
```
madkinsz, line 599: pylint would say not to use the `elif` for clarity after an exit (ie `continue`) -- also makes the diff a little easier but I'm not sure I care.

```suggestion
        if event.parameter_defaults or event.labels is not None:
```

src/prefect_server/api/flows.py
```diff
@@ -579,19 +579,31 @@ async def schedule_flow_runs(flow_id: str, max_runs: int = None) -> List[str]:
     # schedule every event with an idempotent flow run
     for event in flow_schedule.next(n=max_runs, return_events=True):
 
+        # TODO: This has potential to miss scheduling runs that all need to be scheduled at the same time
```
madkinsz, line 582: Can we link this todo to a github issue?

madkinsz (commented):
I can't weigh in on the implications of the TODO, only some minor comments

src/prefect_server/api/flows.py
joshmeek: Yeah I can open a followup issue 👍

joshmeek: I don't have an issue yet to link it to, this is more of a visual edge case from looking at the code where I can't prove it happening (and don't have a reproducible example) but the assumption holds true

cicdw: @joshmeek I believe this should do the trick:
```python
clocks = [
    prefect.schedules.clocks.IntervalClock(
        start_date=now.add(minutes=1),
        interval=datetime.timedelta(minutes=2),
        parameter_defaults=dict(x=i),
    )
    for i in range(20)
]

flow = prefect.Flow(
    tasks=[Parameter("x", default=None)],
    name="Test Scheduled Flow",
    schedule=prefect.schedules.Schedule(clocks=clocks),
)
flow_id = await api.flows.create_flow(
    project_id=project_id, serialized_flow=flow.serialize()
)
await models.FlowRun.where({"flow_id": {"_eq": flow_id}}).delete()
assert len(set((await api.flows.schedule_flow_runs(flow_id)))) == 10
assert len(set((await api.flows.schedule_flow_runs(flow_id)))) == 10 # this will fail in your implementation
```

cicdw: errr sorry that might not be correct, it might still schedule 10 runs just not the correct runs - you'll need to test for all 20 runs with the correct parameter values

tests/api/test_flows.py
```diff
@@ -1308,6 +1308,54 @@ async def test_allows_for_same_time_if_event_attrs_are_different(
...
+        )
+        flow.add_task(prefect.Parameter("x", default=1))
+        flow_id = await api.flows.create_flow(
+            project_id=project_id, serialized_flow=flow.serialize()
+        )
+        await models.FlowRun.where({"flow_id": {"_eq": flow_id}}).delete()
+        assert len(set((await api.flows.schedule_flow_runs(flow_id)))) == 10
+
+        await models.FlowRun.where({"flow_id": {"_eq": flow_id}}).update(
+            set=dict(idempotency_key=None)
```
cicdw, line 1355: I gave this some thought and think we should:

- remove this piece from the test where we reset the idempotency key
- remove [this line](https://github.com/PrefectHQ/server/blob/master/src/prefect_server/api/runs.py#L53) so that idempotency keys are global and non-expiring; right the 24 hour window is arbitrary and users could easily control this when sending keys themselves by sending appropriately rounded timestamps

cicdw (changes requested):
Request to take this PR one step further

src/prefect_server/api/flows.py
joshmeek: @cicdw A bit confused, are you saying we should make a test where we have the expectation that it will fail since it will in the current implementation?

cicdw: No, I was suggesting a reproducible example for an issue

cicdw (approved):
Awesome - this LGTM!

PR #136, opened 2020-11-13
Add a KubernetesAgent to the Helm chart

madkinsz:
## Summary

Adds a KubernetesAgent deployment to the Helm chart

## Importance

Most users deploying Server on K8s will want to use the K8s agent as well.

## Checklist

This PR:


helm/prefect-server/README.md
```diff
@@ -57,3 +63,23 @@ This endpoint must be accessible by *the user* of the UI, not just the server, s
...
+
+query {
+    mutation($input: get_runs_in_queue_input!) {
+            get_runs_in_queue(input: $input) {
+                flow_run_ids
+        }
+    }
+}
+variables {
+    {"input": {"before": "2020-11-13T19:16:01.973734+00:00", "labels": ["n", "u", "l", "l"], "tenant_id": null}}
```
cicdw, line 83: Does `null` / `None` really get parsed like this into individual characters?  If so, that seems like a bug?  The default should be `"labels": []`

helm/prefect-server/templates/agent/deployment.yaml
```diff
@@ -0,0 +1,91 @@
...
+        {{- end }}
+        image: "{{ .Values.agent.image.name }}:{{ .Values.agent.image.tag |  default .Values.prefectVersionTag }}"
+        imagePullPolicy: {{ .Values.agent.image.pullPolicy  }}
+        command:
+          - bash
+          - "-c"
+          - "prefect agent kubernetes start"
+        env:
+          - name: PREFECT__CLOUD__AGENT__AUTH_TOKEN
+            value: ''
```
cicdw, line 43: Is this env var necessary?  It's not strictly speaking wrong, but it might be confusing given that it doesn't have any effect w/ server

helm/prefect-server/templates/apollo/deployment.yaml
```diff
@@ -61,7 +61,7 @@ spec:
             httpGet:
               path: /.well-known/apollo/server-health
               port: http
-          {{- with .Values.apollo.resources -}}
+          {{- with .Values.apollo.resources }}
```
cicdw, line 64: For my own knowledge, what are these changes for?  Seems to be removing a closing `-`?

cicdw (changes requested):
Request to remove one unnecessary env var for clarity, and a few minor questions but overall this LGTM!

helm/prefect-server/templates/apollo/deployment.yaml
madkinsz: When `apollo.resources` is not defined, that whole block is empty as intended. When `apollo.resources` was defined, it was templated in but the `-` strips whitespace so we ended up with

```
              port: httpresources:
              limits: 
                cpu: 12
```
removing the trailing `-` doesn't strip the newline/indent before `resources:`

I'd like to write some python tests to ensure everything optional templates in correctly at some point.

helm/prefect-server/templates/agent/deployment.yaml
madkinsz: I'm happy to remove it, I did the minimal changes necessary from the CLI generated yaml.

helm/prefect-server/README.md
madkinsz: I agree the default should be changed. This must be the intended behavior of toJson 🤷 weird though.

Is this what we want?
```
       - name: PREFECT__CLOUD__AGENT__LABELS
          value: "[\"foo\",\"bar\"]"
```

madkinsz: (good catch btw)

cicdw: Classic javscript; yup `"[\"foo\",\"bar\"]"` looks good and `"[]"` as default is correct 👍

helm/prefect-server/templates/apollo/deployment.yaml
cicdw: Ahh interesting, makes sense - thanks!

cicdw (approved):
Nice, LGTM!

PR #137, opened 2020-11-13
Allow the scheduler to generate names based on the flow run name template

znicholasbrown:
## Summary

Utilizes the new `flow_run_name_template` on the `Clock` object to auto-generate flow run names based on parameter defaults. 

Blocked by: https://github.com/PrefectHQ/prefect/pull/3664

## Importance

Flexibility for users!

## Checklist

This PR:

### DIFF
--- a/src/prefect_server/api/flows.py
+++ b/src/prefect_server/api/flows.py
@@ -550,7 +550,6 @@ async def schedule_flow_runs(flow_id: str, max_runs: int = None) -> List[str]:
         apply_schema=False,
     )
 
-
     if not flow:
         logger.debug(f"Flow {flow_id} can not be scheduled.")
         return run_ids
--- a/src/prefect_server/api/flows.py
+++ b/src/prefect_server/api/flows.py
@@ -601,7 +601,11 @@ async def schedule_flow_runs(flow_id: str, max_runs: int = None) -> List[str]:
             parameters=event.parameter_defaults,
             labels=event.labels,
             idempotency_key=idempotency_key,
-            flow_run_name=event.flow_run_name_template.format(**dict(event.parameter_defaults)) if event.flow_run_name_template is not None else None
+            flow_run_name=event.flow_run_name_template.format(
+                **dict(event.parameter_defaults)
+            )
+            if event.flow_run_name_template is not None
+            else None,
         )
 
         logger.debug(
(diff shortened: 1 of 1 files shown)


src/prefect_server/api/flows.py
```diff
@@ -601,6 +601,11 @@ async def schedule_flow_runs(flow_id: str, max_runs: int = None) -> List[str]:
             parameters=event.parameter_defaults,
             labels=event.labels,
             idempotency_key=idempotency_key,
+            flow_run_name=event.flow_run_name_template.format(
+                **dict(event.parameter_defaults)
+            )
+            if event.flow_run_name_template is not None
+            else None,
```
madkinsz, line 608: As written, we will allow `flow_run_name_template` to be `""` generating empty flow run names. I don't think this is handled in Core either. It may be worth just saying

```suggestion
            flow_run_name=event.flow_run_name_template.format(
                **dict(event.parameter_defaults)
            )
            if event.flow_run_name_template else None,
```

madkinsz (changes requested):

src/prefect_server/api/flows.py
cicdw: Good catch @madkinsz - if we do decide to perform string templating, this should be the property of the Clock / Schedule and not the scheduler IMO, which would mean that this logic should move to Core.

sashgorokhov:
Can we also include 

```python
        flow_run_name = None
        if event.flow_run_name_template is not None:
            formatting_kwargs = {
                'flow_id': flow_id,
                'flow_name': slugify.slugify(flow.name),
                'scheduled_start_datetime': event.start_time,
                'scheduled_start_date': event.start_time.date(),
                'scheduled_start_time': event.start_time.time(),
            }
            flow_run_name = event.flow_run_name_template.format(**formatting_kwargs)
```
https://github.com/sashgorokhov-forks/server/tree/schedule-templated-flow-run-names

This can allow to set run names like `daily-{scheduled_start_date}` and can be easy to distinguish from manually created runs.

I was doing my own impl of this feature but looks like you guys are too fast :)

cicdw:
Closing as stale

MariusHull:
Has this been abandoned ? Would love to get updates on this issue !

madkinsz:
> Has this been abandoned ? Would love to get updates on this issue !

This proved more complex than we thought and will likely be addressed in Prefect 2 but not Prefect 1.

PR #139, opened 2020-11-17
Fix handling for managed postgres usernames with `@` signs

madkinsz:
## Summary

Fixes database migration failures for Azure managed Postgres due to connection string issues.

Azure Postgres instances have usernames "user@instance-name" which requires escaping as "user%40instance-name" for Hasura (and other database connections) to work correctly. However, for migrations, an `@` is required in place of the `%40`

## Checklist

This PR:


cicdw (approved):
Awesome, good catch

PR #140, opened 2020-11-18
Remove `database_url` from 'server database upgrade'

madkinsz:
## Summary

Removes the unused `database_url` option which does not have any effect because the database connstr is loaded in `services/postgres/alembic/env.py`

## Importance

Misleading to have it but not have it work. Dangerous for migrations.

## Checklist

This PR:


madkinsz:
We could also fix this functionality but I'm not sure it's ever going to be used and I'm not sure how changing the config temporarily will work with `alembic.command`

jlowin:
I think that makes sense - this will have to be set via env var

cicdw (approved):

PR #144, opened 2020-12-02
Add shutdown flag for loop services

cicdw:
## Summary

Unused, intended for downstream use cases.

## Checklist

This PR:

### DIFF
--- a/src/prefect_server/api/runs.py
+++ b/src/prefect_server/api/runs.py
@@ -334,7 +334,7 @@ async def get_or_create_task_run_info(
         id=run.returning.id,
         version=db_task_run.version,
         state="Pending",
-        serialized_state=db_task_run.serialized_state,
+        serialized_state=db_task_run_state.serialized_state,
     )
 
 
--- a/src/prefect_server/services/loop_service.py
+++ b/src/prefect_server/services/loop_service.py
@@ -19,6 +19,9 @@ class LoopService:
     # if no loop_seconds_config_key is provided, this will be the default
     loop_seconds_default = 600
 
+    # shutdown flag for gracefully exiting the infinite loop
+    is_running = True
+
     def __init__(self, loop_seconds: Union[float, int] = None):
         if loop_seconds is None:
             if self.loop_seconds_config_key:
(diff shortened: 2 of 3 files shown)


jlowin (approved):

src/prefect_server/api/runs.py
```diff
@@ -334,7 +334,7 @@ async def get_or_create_task_run_info(
         id=run.returning.id,
         version=db_task_run.version,
         state="Pending",
-        serialized_state=db_task_run.serialized_state,
+        serialized_state=db_task_run_state.serialized_state,
```
madkinsz, line 337: Find this mistake too?

src/prefect_server/services/loop_service.py
```diff
@@ -19,6 +19,9 @@ class LoopService:
     # if no loop_seconds_config_key is provided, this will be the default
     loop_seconds_default = 600
 
+    # shutdown flag for gracefully exiting the infinite loop
+    is_running = True
```
madkinsz, line 23: Does it make sense to have this be at the class level?

e.g.

```
In [1]: class foo:
   ...:     x = True
   ...: 

In [2]: foo.x
Out[2]: True

In [3]: instance = foo()

In [4]: instance.x
Out[4]: True

In [5]: foo.x = False

In [6]: instance.x
Out[6]: False
```

src/prefect_server/api/runs.py
cicdw: Yup, fixes an issue - I'll rebuild the relevant tagged server images after this is merged.

src/prefect_server/services/loop_service.py
cicdw: Great question - all services which run in the same process will shutdown simultaneously either way (there's no need to selectively shut down a single service), so it's not necessary to make it an instance variable.

madkinsz (approved):

src/prefect_server/services/loop_service.py
madkinsz: Might be worth noting in a comment but that makes sense!

PR #149, opened 2020-12-02
Add index on task run names for global search

madkinsz:
## Summary

Corresponding to https://github.com/PrefectHQ/cloud/pull/2884 adds a GIN trigram text search index on task run names

## Importance

Necessary for reasonable global search of task run names

## Checklist

This PR:

### DIFF
--- a/services/postgres/alembic/versions/2020-12-02T130956_add_index_for_task_run_names.py
+++ b/services/postgres/alembic/versions/2020-12-02T130956_add_index_for_task_run_names.py
@@ -19,12 +19,15 @@ depends_on = None
 
 
 def upgrade():
-    op.create_index(
-        "ix_task_run_name",
-        "task_run",
-        ["name"],
+    op.execute(
+        """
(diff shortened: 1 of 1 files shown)


services/postgres/alembic/versions/2020-12-02T130956_add_index_for_task_run_names.py
```diff
@@ -0,0 +1,33 @@
...
+revision = "57ac2cb01ac1"
+down_revision = "3c87ad7e0b71"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.execute(
+        """
+        CREATE INDEX
```
madkinsz, line 24: @jlowin Do we do concurrent index creation in `Server` .. ?

jlowin: I don't think it's necessary here

jlowin (approved):

PR #151, opened 2020-12-03
Always pull prefecthq docker images in the helm chart

zanieb:
## Summary

Changes the default setting from IfNotPresent to Always since we are
using the 'latest' tag. This matches K8s behavior if the pull policy
were null. I considered setting this to null by default and depending
on the K8s default handling depending on the tag value, but it sounds
like the 'latest' -> 'Always' default pull policy behavior is
deprecated and I don't want to rely on it.

We will likely set this back to IfNotPresent when we pin the
prefectVersion for helm chart releases. However, some people prefer
Always because then the image cannot accidentally be grabbed by a
deployment without the proper secrets (not relevant for our images
since they are public)

Important: This does introduce the issue that if a deployment is rolled,
it could pull a different version than the other deployments. This is
not ideal for some users as it introduces version compatibility issues.
I'll add a note about versioning to the README suggesting pinning
`prefectVersion` for production deployments.

Edit: See [rendered readme section](https://github.com/PrefectHQ/server/blob/helm-always-pull/helm/prefect-server/README.md#versioning)

## Importance

Installations with this chart will be stuck on old versions otherwise.

## Checklist

This PR:


cicdw:
I know this is still a WIP but your reasoning makes sense to me ✅

cicdw (approved):

PR #157, opened 2020-12-09
Add events resource to Kubernetes Agent RBAC helm chart template

joshmeek:
The next release of core will allow agents to report events for pending pods that belong to jobs it creates. (PrefectHQ/prefect#3783) In order to use this functionality the in-cluster agent in the helm chart should have the proper resources in the RBAC.


madkinsz (approved):
Sorry about the delay!

PR #158, opened 2020-12-11
Task run heartbeats refactor

cicdw:
## Summary

This PR updates the zombie killer to rely on _flow_ run heartbeats instead of individual task run heartbeats; we've found that task run heartbeat processes can be brittle and cause resource contention issues when working with thousands of parallel runs.  In addition, if the Flow Run process dies then the task runs it submits also die, so this change should not affect behavior.

## Importance

Performance improvement!

## Checklist

This PR:

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -5,7 +5,7 @@ from setuptools import find_packages, setup
 import versioneer
 
 install_requires = [
-    "prefect >= 0.13.0",
+    "prefect < 0.14.0",  # temporary
     "alembic >= 1.2, < 2.0",
     "ariadne >= 0.8.0, < 0.12.0",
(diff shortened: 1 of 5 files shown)


setup.py
```diff
@@ -5,7 +5,7 @@
 import versioneer
 
 install_requires = [
-    "prefect >= 0.13.0",
+    "prefect < 0.14.0",  # temporary
```
cicdw, line 8: This is just to avoid test issues as the 0.14.0 changes are merged, some of which remove old objects that are referenced in tests; we can do a quick audit prior to release and unbound this

jlowin (approved):

PR #163, opened 2020-12-13
Remove .get() and .exists() methods on hasura client

### LINKED ISSUE #160: Explore whether the Hasura `.get()` and `.exists()` methods can be safely removed
## Current behavior

These methods are cruft that, if used, can be easily replaced with "proper" queries. https://github.com/PrefectHQ/server/blob/master/src/prefect_server/database/hasura.py#L170
https://github.com/PrefectHQ/server/blob/master/src/prefect_server/database/hasura.py#L186

In particular, they rely on a `by_pk` query that we want to avoid.

Note they are referenced in `orm.py` and possibly in the codebase.

jlowin:
## Summary

Closes #160 

Removes `.get()` and `.exists()` from the Hasura client (both brittle fields that depended on `by_pk` queries and were almost unused), as well as `.exists()` from the ORM, as it is easily replaced with `.where(id=id).first()`.

## Importance

Removing brittle methods and dependencies on Hasura implementation details (like `by_pk`)

## Checklist

This PR:

### DIFF
--- a/src/prefect_server/database/hasura.py
+++ b/src/prefect_server/database/hasura.py
@@ -167,29 +167,6 @@ class HasuraClient(GraphQLClient):
 
         return result
 
-    async def get(
-        self, graphql_type: str, id: str, selection_set: GQLObjectTypes
-    ) -> Box:
-        """
-        Query a specific object type by ID
-
-        Args:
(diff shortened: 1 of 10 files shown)


cicdw (approved):
LGTM

PR #166, opened 2020-12-28
Per flow-run run-configs

jcrist:
This adds support for configuring a `run_config` at both the flow-run and flow-group levels, allowing for customization of a flow's deployment environment without re-registration (and on a per-run basis). The following changes were required:

- Addition of a `run_config` column in the `flow_group` table.  This is `null` by default, only filled in if a `run_config` is set on the flow group.
- Addition of a `run_config` column in the `flow_run` table. This is (almost) never null (unless `flow.run_config` is null, which *is* supported) - if no override is provided, the `run_config` on the flow is copied over. This means that `flow_run.run_config` is the source of truth for a flow-run's run-config. The agents will be updated accordingly.
- Addition of a `set_flow_group_run_config` graphql mutation for setting a flow-group's `run_config`. As with flow group labels, this can be set as `None` to clear the override.
- Addition of an optional `run_config` field in the `create_flow_run` graphql mutation.

A few things a reviewer should note:

- I did not add a `set_flow_run_run_config` mutation, unlike per-run labels, per-run `run_config` can only be created at creation of the run, not updated after creation. I went with this because updating a run_config afterwards might need to trigger updates to the run's labels as well, which is tricky. We can always add this route later if needed.
- A flow run's labels are computed with the following precedence:
  - `labels` explicitly passed to `create_flow_run`
  - `flow_group.labels`, if present
  - `run_config.labels` on a `run_config` explicitly passed to `create_flow_run`
  - `flow_group.run_config.labels`, if present
  - `flow.run_config.labels`, if present (only not present if legacy `flow.environment` is used).
- Since `flow_group.run_config.labels` and `flow_group.labels` now both exist, the label logic in the flow UI views will need to be updated as well.

Part of https://github.com/PrefectHQ/prefect/issues/3858 (needs cloud, core, and UI PRs as well).

This PR:

### DIFF
--- a/services/postgres/alembic/versions/2020-12-28T150027_add_run_config_to_flow_runs.py
+++ b/services/postgres/alembic/versions/2020-12-28T150027_add_run_config_to_flow_runs.py
@@ -12,8 +12,8 @@ from sqlalchemy.dialects.postgresql import JSONB
 
 
 # revision identifiers, used by Alembic.
-revision = '7ca57ea2fdff'
-down_revision = '57ac2cb01ac1'
+revision = "7ca57ea2fdff"
+down_revision = "57ac2cb01ac1"
 branch_labels = None
 depends_on = None
 
--- a/services/postgres/alembic/versions/2020-12-28T150027_add_run_config_to_flow_runs.py
+++ b/services/postgres/alembic/versions/2020-12-28T150027_add_run_config_to_flow_runs.py
@@ -25,6 +25,24 @@ def upgrade():
     op.add_column(
         "flow_group", sa.Column("run_config", JSONB, nullable=True, server_default=None)
     )
+    # This query updates all currently scheduled flow runs to add the run_config field
+    # from the corresponding flow.
+    op.execute(
+        """
+        WITH runs_to_update AS (
+            SELECT
+                flow_run.id,
+                flow.run_config
+            from flow_run
+            join flow on flow_run.flow_id = flow.id
+            where flow_run.state = 'Scheduled' OR flow_run.updated > NOW() - interval '1 day'
+        )
+        update flow_run
+        set run_config = runs_to_update.run_config
+        from runs_to_update
+        where flow_run.id = runs_to_update.id;
+        """
+    )
 
 
 def downgrade():
--- a/services/postgres/alembic/versions/2020-12-28T150027_add_run_config_to_flow_runs.py
+++ b/services/postgres/alembic/versions/2020-12-28T150027_add_run_config_to_flow_runs.py
@@ -1,0 +1,32 @@
+"""
+Add run_config to flow runs and flow groups
+
+Revision ID: 7ca57ea2fdff
+Revises: 57ac2cb01ac1
+Create Date: 2020-12-28 15:00:27.573816
+
+"""
+from alembic import op
+import sqlalchemy as sa
+from sqlalchemy.dialects.postgresql import JSONB
+
+
+# revision identifiers, used by Alembic.
+revision = '7ca57ea2fdff'
+down_revision = '57ac2cb01ac1'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.add_column(
+        "flow_run", sa.Column("run_config", JSONB, nullable=True, server_default=None)
+    )
+    op.add_column(
+        "flow_group", sa.Column("run_config", JSONB, nullable=True, server_default=None)
+    )
+
+
+def downgrade():
+    op.drop_column("flow_run", "run_config")
+    op.drop_column("flow_group", "run_config")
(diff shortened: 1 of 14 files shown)


jcrist:
Test failure is unrelated.

services/postgres/alembic/versions/2020-12-28T150027_add_run_config_to_flow_runs.py
```diff
@@ -0,0 +1,50 @@
...
+                flow_run.id,
+                flow.run_config
+            from flow_run
+            join flow on flow_run.flow_id = flow.id
+            where flow_run.state = 'Scheduled' OR flow_run.updated > NOW() - interval '1 day'
+        )
+        update flow_run
+        set run_config = runs_to_update.run_config
+        from runs_to_update
+        where flow_run.id = runs_to_update.id;
```
jcrist, line 43: I confess I don't really know SQL, there's a chance this is wrong or could be done better.

src/prefect_server/api/flow_groups.py
```diff
@@ -159,6 +159,33 @@ async def set_flow_group_labels(flow_group_id: str, labels: List[str] = None) ->
...
+
+    Returns:
+        - bool: whether setting `run_config` for the flow group was successful
+
+    Raises:
+        - ValueError: if flow group ID isn't provided
+    """
+    if not flow_group_id:
+        raise ValueError("Invalid flow group ID")
+    result = await models.FlowGroup.where(id=flow_group_id).update(
```
cicdw, line 183: Should we perform any validation on this input (e.g., attempt to deserialize it)?

src/prefect_server/api/runs.py
```diff
@@ -173,13 +182,20 @@ async def _create_flow_run(
...
+            run_config = flow.run_config
+
     # set labels
     if labels is not None:
         run_labels = labels
     elif flow.flow_group.labels is not None:
         run_labels = flow.flow_group.labels
-    elif flow.run_config is not None:
-        run_labels = flow.run_config.get("labels") or []
+    elif run_config is not None:
```
cicdw, line 197: I think the logic in this section needs to be rearranged; specifically if a user provides a run config (with labels) but forgets to redundantly supply the labels in the API call, then the labels will be pulled from the flow group instead of the more recent / immediate run config object.  

One way we could achieve this is via an `else:` statement on the `if run_config is None:` clause above that does something like:
```python
elif labels is None:
    labels = run_config.get("labels")
```

and then this `elif` could be removed entirely.

src/prefect_server/graphql/schema/flows.graphql
```diff
@@ -151,7 +154,14 @@ input set_flow_group_labels_input {
   labels: [String!]
 }
 
-	input set_flow_group_schedule_input {
+input set_flow_group_run_config_input {
+  "The ID of the flow group to update"
+  flow_group_id: UUID!
+  "The serialized run config to associate with the flow group"
+  run_config: JSON
```
cicdw, line 161: Confirming that `run_config` is optional because we want to allow `None` to be provided - is that accurate?

src/prefect_server/graphql/flow_groups.py
```diff
@@ -26,6 +26,16 @@ async def resolve_set_flow_group_labels(
     return {"success": result}
 
 
+@mutation.field("set_flow_group_run_config")
+async def resolve_set_flow_group_run_config(
+    obj: Any, info: GraphQLResolveInfo, input: dict
+) -> dict:
+    result = await api.flow_groups.set_flow_group_run_config(
+        flow_group_id=input["flow_group_id"], run_config=input["run_config"]
```
cicdw, line 34: if `run_config` is intended to be optional, we'll need to account for that here
```suggestion
        flow_group_id=input["flow_group_id"], run_config=input.get("run_config")
```

cicdw (changes requested):
Good stuff - some minor questions and feedback

src/prefect_server/graphql/schema/flows.graphql
jcrist: Correct.

src/prefect_server/api/flow_groups.py
jcrist: Cloud/server don't have models for run configs. We could add one, but it'd just check for the presence of a `type` and `labels` field (which is optional). Could go either way on this.

src/prefect_server/api/runs.py
jcrist: Hmmm, I thought about that, but decided the order I described above made more sense to me. Happy to swap though, I can see arguments for both ways.

jcrist: Swapped the resolution order (this wasn't a bug per-se, just a different idea of what was expected behavior).

jcrist:
I believe all comments have been addressed.

src/prefect_server/api/flow_groups.py
cicdw: Core technically has a serializer but I suppose that would destroy the benefits of using pydantic for compatibility purposes; we don't have to do anything about this now, just something to keep in the back of our minds

cicdw (approved):

PR #169, opened 2020-12-30
Flow Group schedule timezones

cicdw:
## Summary

This PR exposes a new `timezone` input to the `set_flow_group_schedule` GraphQL mutation.  Allowable timezones can be obtained via `pendulum.timezones` in Python (**cc:** @znicholasbrown / @ThatGalNatalie in case you need to audit this list and ensure the UI can serve these values appropriately).

For learning purposes, this PR demonstrates how the lifecycle of an API request works: clients send JSON payloads corresponding to GraphQL inputs.  The GraphQL resolver layer then manipulates / handles those, and passes them deeper into the API function layer.  Here, additional manipulation / validation may occur before placing the payloads into the database.  It is usually much easier for testing purposes to contain most of the logic / validation in the API layer.

In this case, the user-facing change is that a convenient / easy to understand `timezone` field is exposed.  Under the hood, within the API layer, we take this timezone and convert it into a `start_date` that can be interpreted by [the relevant Core Schedule serializers](https://github.com/PrefectHQ/prefect/blob/master/src/prefect/serialization/schedule.py) for the scheduler to consume / work with.  **NB:** timezones are specified via `start_date`s on schedules

## Importance

Server-solution to https://github.com/PrefectHQ/ui/issues/423

## Checklist

This PR:

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,5 +1,13 @@
 # Changelog
 
+## December 16, 2020 <Badge text="beta" type="success" />
+
+Released on December 16, 2020.
+
+### Enhancements
+
+- Add the `events` RBAC permission to agents created by the helm chart - [#157](https://github.com/PrefectHQ/cloud/pull/157)
+
 ## December 14, 2020 <Badge text="beta" type="success" />
 
 Released on December 14, 2020.
--- a/changes/pr157.yaml
+++ b/changes/pr157.yaml
@@ -1,2 +1,0 @@
-enhancement:
-  - "Add the `events` RBAC permission to agents created by the helm chart - [#157](https://github.com/PrefectHQ/cloud/pull/157)"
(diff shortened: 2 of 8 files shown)


Changelog.md
```diff
@@ -1,5 +1,13 @@
 # Changelog
 
+## December 16, 2020 <Badge text="beta" type="success" />
+
+Released on December 16, 2020.
+
+### Enhancements
+
+- Add the `events` RBAC permission to agents created by the helm chart - [#157](https://github.com/PrefectHQ/cloud/pull/157)
+
```
madkinsz, line 10: How'd this get in here?

madkinsz (approved):

madkinsz:
Once the changelog diff is fixed lgtm!

Changelog.md
cicdw: I accidentally cut the changelog for [a release from that day](https://github.com/PrefectHQ/server/releases/tag/2020-12-16) locally, but never pushed up my changes 😬

PR #176, opened 2021-01-19
Fix tests in response to some Core changes

cicdw:
Minor interface change https://github.com/PrefectHQ/prefect/pull/3969 caused some of our service tests to fail


madkinsz (approved):
Easy enough fix

PR #177, opened 2021-01-20
First class descriptions

cicdw:
## Summary

This PR adds formal support for Flow Group descriptions that can be edited via API.  In particular, we now have the following ways of creating and managing descriptions for flows and flow groups:
- passing a `description` when calling `mutation{create_flow}` (this has been supported for a long time -- *cc* @jcrist in case you wanted to begin using this in Core)
- calling the `set_flow_group_description` mutation (this is what this PR adds)

Note that this flow group field will be treated _slightly_ differently from other flow group fields in that the description will not be an override.  Instead, it will be editable and persistent, but individual versions may still have their own descriptions displayed (but note that they won't be editable).

All descriptions will support markdown formatting when displayed.

**cc:** @zhen0 

## Importance

This will enable a more fully featured UI experience for flows, especially for those that 

## Checklist

This PR:

### DIFF
--- a/src/prefect_server/api/flow_groups.py
+++ b/src/prefect_server/api/flow_groups.py
@@ -182,7 +182,7 @@ async def set_flow_group_description(
 
     Args:
         - flow_group_id (str): the ID of the flow group to update
-        - description (dict, optional): A markdown-compatible description for this flow group.
+        - description (str, optional): A markdown-compatible description for this flow group.
             Providing `None` resets any previous flow group
             `description` setting.
 
--- a/src/prefect_server/api/flow_groups.py
+++ b/src/prefect_server/api/flow_groups.py
@@ -173,6 +173,33 @@ async def set_flow_group_labels(flow_group_id: str, labels: List[str] = None) ->
     return bool(result.affected_rows)
 
 
+@register_api("flow_groups.set_flow_group_description")
+async def set_flow_group_description(
+    flow_group_id: str, description: str = None
+) -> bool:
+    """
+    Sets description for a flow group.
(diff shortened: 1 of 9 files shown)


src/prefect_server/database/models.py
```diff
@@ -307,6 +307,7 @@ class FlowGroup(HasuraModel):
     id: UUIDString = None
     created: datetime.datetime = None
     updated: datetime.datetime = None
+    description: str = None
```
madkinsz, line 310: I'm confused that it's a string here but a json blob in the database and graphql schema

src/prefect_server/api/flow_groups.py
```diff
@@ -173,6 +173,33 @@ async def set_flow_group_labels(flow_group_id: str, labels: List[str] = None) ->
...
+@register_api("flow_groups.set_flow_group_description")
+async def set_flow_group_description(
+    flow_group_id: str, description: str = None
+) -> bool:
+    """
+    Sets description for a flow group.
+
+    Args:
+        - flow_group_id (str): the ID of the flow group to update
+        - description (dict, optional): A markdown-compatible description for this flow group.
```
madkinsz, line 185: `dict` here but `str` in the type hint

src/prefect_server/database/models.py
cicdw: You are correct to be confused - it should be a string everywhere; will update!

cicdw: Updated

madkinsz (approved):

PR #178, opened 2021-01-26
Update tag versioning scheme to be PEP 440 compliant

cicdw:
We began seeing some build errors recently (interestingly enough, they appeared to be non-deterministic) because Prefect Server's versioning scheme was not [PEP 440](https://www.python.org/dev/peps/pep-0440) compliant, resulting in invalid wheels.

This PR updates our jobs to detect versions based on `YYYY.MM.DD` instead of `YYYY-MM-DD`

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -221,7 +221,7 @@ workflows:
             branches:
               ignore: /.*/
             tags:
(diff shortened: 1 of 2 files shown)


jlowin (dismissed):

madkinsz (dismissed):
Huh. Alrighty then. Might want a changelog entry for this since it is user-facing in a way.

madkinsz (approved):

PR #180, opened 2021-01-31
Annotate types as optional

dsaxton:
## Summary

Hi all, typing nitpick here but I noticed there are a lot of optional types that are not annotated as such so I thought I'd start doing that. This doesn't cover them all, but if this is a change you would like to make I can fix some more (some function signatures are only partially fixed in this PR at the moment, for example).

## Importance

Would allow for better inferences by tools like mypy.

## Checklist

This PR:

### DIFF
--- a/src/prefect_server/api/agents.py
+++ b/src/prefect_server/api/agents.py
@@ -1,4 +1,4 @@
-from typing import List
+from typing import List, Optional
 
 import pendulum
 
@@ -11,9 +11,9 @@ from prefect_server.utilities import context
 async def register_agent(
     tenant_id: str,
     labels: List[str],
-    agent_config_id: str = None,
-    name: str = None,
-    type: str = None,
+    agent_config_id: Optional[str] = None,
+    name: Optional[str] = None,
+    type: Optional[str] = None,
 ) -> str:
     """
     Register a new agent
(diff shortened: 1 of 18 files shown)


jlowin:
Mypy automatically infers the `Optional` typing when a default is `None`, but I just learned that behavior is being deprecated: https://github.com/python/mypy/issues/9091

dsaxton:
> Mypy automatically infers the `Optional` typing when a default is `None`, but I just learned that behavior is being deprecated: [python/mypy#9091](https://github.com/python/mypy/issues/9091)

Ah, I didn't know that, thanks! I'll close then, since this is a lot of extra boilerplate.

jlowin:
Sorry @dsaxton, I didn't mean to write that to shut down the PR - I was going to share it with you in case you didn't know, but then learned it's now an unsupported behavior! I think it raises a good question of the degree to which we want to support mypy, as currently Server does not enforce mypy semantics (though Core does).

dsaxton:
> Sorry @dsaxton, I didn't mean to write that to shut down the PR - I was going to share it with you in case you didn't know, but then learned it's now an unsupported behavior! I think it raises a good question of the degree to which we want to support mypy, as currently Server does not enforce mypy semantics (though Core does).

No worries, let me know if you'd like to reopen. It does seem though that relying instead on the `--implicit-optional` flag may be the easier / preferred approach (depending on how pedantic we want to be about what the "correct" type is).

PR #183, opened 2021-02-04
Helm: Optionally create a tenant

michcio1234:
## Summary
In Helm chart, add an option to automatically create a default tenant.

## Importance
It's a bit cumbersome to configure a local CLI client to connect to a custom Kubernetes deployment, especially when there are no flags like `--host` and `--port`. For an inexperienced user, like me, it was far from obvious how to do it. Yet this is the only way I found in the docs to create a tenant. I thought it would be good if the tenant was added automatically.

## Misc
I couldn't find a way of checking whether a tenant is already created, that's why I fell back to just trying to create it, and checking for a particular error to see if it's already there. I know it's ugly.

I am just beginning getting to know Prefect. I am sure that there is a prettier way of achieving what I've done here. Yet this is what I was able to come up with and I suppose it would be helpful for others until someone does it in a more elegant way.

## Checklist
This PR:

- [x] adds new tests (doesn't apply, I guess, but I tested it as well as I could locally)

### DIFF
--- a/helm/prefect-server/README.md
+++ b/helm/prefect-server/README.md
@@ -39,11 +39,14 @@ Export the password as the error asks then set it within the subchart using `--s
 See comments in `values.yaml`.
 
 ### Tenant
+A tenant is needed for the server API to be operational.
+If there is no tenant in the database, the UI dashboard will fail to display and agents will error.
 
-To automatically create a default tenant, use the flag `--set jobs.createTenant.enabled=true`. 
-A Tenant is needed for the dashboard and the agent to run properly.
+To automatically create a default tenant, use the flag `--set jobs.createTenant.enabled=true`.
 
-Alternatively, you can create a tenant manually. See [Troubleshooting](#troubleshooting) section.
+To create the tenant manually when your chart is already installed, 
+run `prefect backend server && prefect server create-tenant --name default --slug default` to create a default tenant.
+Check out [Connecting to your Server](#connecting-to-your-server) to connect your local `prefect` to your server before creating the tenant.
 
 ### KubernetesAgent
 
@@ -103,12 +106,7 @@ Alternatively, you can port-forward the apollo service to your localhost
 
 If you go to the 'Home' page, it will likely direct you to create a tenant. Without a tenant, the dashboard cannot display.
 
-See [Tenant](#tenant) section to learn how to automatically create the default tenant during chart
-installation.
-
-To create the tenant manually when your chart is already installed, 
-run `prefect backend server && prefect server create-tenant --name default --slug default` to create a default tenant.
-Check out [Connecting to your Server](#connecting-to-your-server) to connect your local `prefect` to your server before creating the tenant.
+See the [Tenant](#tenant) section to learn how to create a default tenant.
 
 ### The UI loads but cannot connect to the API
 
--- a/helm/prefect-server/README.md
+++ b/helm/prefect-server/README.md
@@ -38,6 +38,12 @@ Export the password as the error asks then set it within the subchart using `--s
 
 See comments in `values.yaml`.
 
+### Tenant
+
+To automatically create a default tenant, use the flag `--set jobs.createTenant.enabled=true`. 
+A Tenant is needed for the dashboard and the agent to run properly.
+
+Alternatively, you can create a tenant manually. See [Troubleshooting](#troubleshooting) section.
 
 ### KubernetesAgent
 
@@ -95,9 +101,13 @@ Alternatively, you can port-forward the apollo service to your localhost
 
 ### The UI loads but the dashboard is blank
 
-If you go to the 'Home' page it will likely direct you to create a tenant. Without a tenant, the dashboard cannot display.
-Run `prefect backend server && prefect server create-tenant --name default --slug default` to create a default tenant.
+If you go to the 'Home' page, it will likely direct you to create a tenant. Without a tenant, the dashboard cannot display.
 
+See [Tenant](#tenant) section to learn how to automatically create the default tenant during chart
+installation.
+
+To create the tenant manually when your chart is already installed, 
+run `prefect backend server && prefect server create-tenant --name default --slug default` to create a default tenant.
 Check out [Connecting to your Server](#connecting-to-your-server) to connect your local `prefect` to your server before creating the tenant.
 
 ### The UI loads but cannot connect to the API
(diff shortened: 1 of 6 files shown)


madkinsz:
Thanks for taking the time to contribute! I think this may be better implemented as an init container for the UI since the agent is an optional component. Otherwise, I think this makes sense. 

I'm a little hesitant as this feels like something we should just put into Server globally rather than patching the creation in the Helm chart / Prefect Core. cc @cicdw on the reason we're not doing that?

michcio1234:
>I think this may be better implemented as an init container for the UI since the agent is an optional component.

Thanks, I'll change it as you're suggesting. But tomorrow (European time). :)

>this feels like something we should just put into Server globally

As much as I don't know Prefect yet, I agree. :)

cicdw:
We have to be careful about automagic configurations like that -- we have had users request that we _not_ automatically create default tenants from Core, hence the creation of the CLI command for creating tenants.  I don't have a strong opinion about the automatic creation in this particular PR, but I do have an opinion that we shouldn't bake it in at the lowest level.

madkinsz:
> We have to be careful about automagic configurations like that -- we have had users request that we _not_ automatically create default tenants from Core, hence the creation of the CLI command for creating tenants. I don't have a strong opinion about the automatic creation in this particular PR, but I do have an opinion that we shouldn't bake it in at the lowest level.

Great thanks Chris -- @michcio1234 then an implementation that creates it from the UI should be fine given that it's a flag in the values file. I don't love that it'd be a UI init container because the UI should be able to start before the API is available but I think since it's off by default that's a reasonable trade-off. I'd ask that you also update the helm chart troubleshooting / readme with references to this config value.

michcio1234:
@madkinsz it looks like UI image doesn't have `prefect` CLI installed. I see three options:
- install `prefect` in UI image - meh
- use a different image, e.g. `prefecthq/prefect` in init container - BUT if UI deployment is scheduled on a node where this image is not present, it implies the need of pulling this image just to create a tenant.
- add an init container not to UI deployment but to agent or other deployment that uses an image in which `prefect` CLI is present

I vote for the last option, but you probably have a better view of what would be optimal.

madkinsz:
Perhaps we should just create a K8s `Job`? I don't really see any reason not to and then it'd only be deployed on helm-chart deploy rather than on every restart of one of our services.

michcio1234:
@madkinsz I made it a Job and updated readme and notes.

michcio1234:
@madkinsz @cicdw @jlowin pinging for someone to have a look. Please let me know if you're okay with merging it. 
To recap: by default nothing changes, but it's possible to create a default tenant using a K8s Job.

helm/prefect-server/templates/NOTES.txt
```diff
@@ -47,5 +47,14 @@
   - The location can also be changed within the UI itself per user
   - The API must be accessible from the the user's machine for the UI to work
 
+{{- if .Values.jobs.createTenant.enabled }}
+
+#4 The `{{ .Values.jobs.createTenant.tenant.name }}` tenant was created.
+   The dashboard and the Agent (if enabled) should become functional without any further action.
+
+{{- else }}
+
```
madkinsz, line 56: I think we can just leave this out. The notes are already a bit lengthy as is and I want to limit it to action items.

helm/prefect-server/README.md
```diff
@@ -38,6 +38,12 @@ Export the password as the error asks then set it within the subchart using `--s
 
 See comments in `values.yaml`.
 
+### Tenant
+
+To automatically create a default tenant, use the flag `--set jobs.createTenant.enabled=true`. 
+A Tenant is needed for the dashboard and the agent to run properly.
```
madkinsz, line 44:
```suggestion
A tenant is needed for the server API to be operational.
```

helm/prefect-server/README.md
```diff
@@ -95,9 +101,13 @@ Alternatively, you can port-forward the apollo service to your localhost
...
 
-If you go to the 'Home' page it will likely direct you to create a tenant. Without a tenant, the dashboard cannot display.
-Run `prefect backend server && prefect server create-tenant --name default --slug default` to create a default tenant.
+If you go to the 'Home' page, it will likely direct you to create a tenant. Without a tenant, the dashboard cannot display.
+
+See [Tenant](#tenant) section to learn how to automatically create the default tenant during chart
+installation.
 
+To create the tenant manually when your chart is already installed, 
+run `prefect backend server && prefect server create-tenant --name default --slug default` to create a default tenant.
```
madkinsz, line 110: Lets move this to the tenant section

helm/prefect-server/README.md
```diff
@@ -95,9 +101,13 @@ Alternatively, you can port-forward the apollo service to your localhost
 
 ### The UI loads but the dashboard is blank
 
-If you go to the 'Home' page it will likely direct you to create a tenant. Without a tenant, the dashboard cannot display.
-Run `prefect backend server && prefect server create-tenant --name default --slug default` to create a default tenant.
+If you go to the 'Home' page, it will likely direct you to create a tenant. Without a tenant, the dashboard cannot display.
+
+See [Tenant](#tenant) section to learn how to automatically create the default tenant during chart
+installation.
```
madkinsz, line 107:
```suggestion
See the [Tenant](#tenant) section to learn how to create a default tenant.
```

helm/prefect-server/README.md
madkinsz: Let's have this line first so

```
A tenant is needed for the server API to be operational.
If there is not a tenant in the database, the UI dashboard will fail to display and agents will error.

To automatically create, ...

To manually create, ...
```

zanieb (commented):
Thanks again, the job itself looks great -- I'm presuming you've tested this locally? A few fixes on the documentation and I'll merge this.

michcio1234:
>I'm presuming you've tested this locally?

Yes, in a minikube cluster.

michcio1234:
@madkinsz updated the docs.

zanieb (approved):
Thanks again @michcio1234 !

michcio1234:
My pleasure, thanks for merging. :)

sm-Fifteen:
Wouldn't it be better to just have the UI offer you to create a tenant if none exist instead of failing to display? All one really needs to do is to connect to the graphql/prefect server endpoint and send

```graphql
 mutation($input: create_tenant_input!) { create_tenant(input: $input) { id } }
```
with parameters
```json
{ "input": { "name": "default", "slug": "default" } }
```

which is exactly what the `prefect server create-tenant` command does. While it's good to have a job to take care of that on server setup for Kubernetes clusters, I'm not entirely sure why that it's done this way instead of being part of the UI's initial setup process.

PR #185, opened 2021-02-05
Better archiving experience

cicdw:
## Summary

Currently, whenever a flow is archived we _delete_ all outstanding runs for that flow that are still in a `Scheduled` state (because they can't run).  This works well for auto scheduled runs; however, users who create ad-hoc runs for a flow that maybe take a while to begin (due to concurrency limits, or maybe an agent malfunction) might register a new version in the meantime.  In this situation, the deletion of these runs can be very mysterious and unexpected.  @znicholasbrown noted that we could instead _cancel_ these runs so that there is a record of what happened.

## Importance

More transparent UX.

## Checklist

This PR:

### DIFF
--- a/src/prefect_server/api/flows.py
+++ b/src/prefect_server/api/flows.py
@@ -372,8 +372,10 @@ async def archive_flow(flow_id: str) -> bool:
         }
     ).get({"id"})
     msg = f"Flow {flow_id} was archived."
+    tasks = []
     for flow_run in fr_ids:
-        await api.states.cancel_flow_run(flow_run.id, state_message=msg)
+        tasks.append(api.states.cancel_flow_run(flow_run.id, state_message=msg))
+    await asyncio.gather(*tasks)
     return True
 
 
--- a/src/prefect_server/api/flows.py
+++ b/src/prefect_server/api/flows.py
@@ -332,7 +332,8 @@ async def archive_flow(flow_id: str) -> bool:
     Archives a flow.
 
     Archiving a flow prevents it from scheduling new runs. It also:
-        - deletes any currently scheduled runs
+        - deletes any currently auto-scheduled runs
+        - cancels any ad-hoc created scheduled runs
         - resets the "last scheduled run time" of any schedules
 
     Args:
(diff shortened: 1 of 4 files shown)


src/prefect_server/api/flows.py
```diff
@@ -355,9 +356,24 @@ async def archive_flow(flow_id: str) -> bool:
...
+    fr_ids = await models.FlowRun.where(
+        {
+            "flow_id": {"_eq": flow_id},
+            "state": {"_eq": "Scheduled"},
+            "auto_scheduled": {"_eq": False},
+        }
+    ).get({"id"})
+    msg = f"Flow {flow_id} was archived."
+    for flow_run in fr_ids:
+        await api.states.cancel_flow_run(flow_run.id, state_message=msg)
```
madkinsz, line 376: Would it make sense to do a `gather` instead in case one fails?

src/prefect_server/api/states.py
```diff
@@ -231,7 +233,7 @@ async def cancel_flow_run(flow_run_id: str) -> models.FlowRun:
         return flow_run
     else:
         if state.is_running():
-            state = Cancelling("Flow run is cancelling")
+            state = Cancelling(state_message or "Flow run is cancelling")
```
madkinsz, line 236: Ahh, cloud only before 👍

madkinsz (commented):
LGTM -- minor comment.

src/prefect_server/api/flows.py
cicdw: Yup, that's a great idea - will update

cicdw: @madkinsz updated 👍

src/prefect_server/api/flows.py
```diff
@@ -355,9 +356,26 @@ async def archive_flow(flow_id: str) -> bool:
...
+            "flow_id": {"_eq": flow_id},
+            "state": {"_eq": "Scheduled"},
+            "auto_scheduled": {"_eq": False},
+        }
+    ).get({"id"})
+    msg = f"Flow {flow_id} was archived."
+    tasks = []
+    for flow_run in fr_ids:
+        tasks.append(api.states.cancel_flow_run(flow_run.id, state_message=msg))
+    await asyncio.gather(*tasks)
```
madkinsz, line 378: I have a preference for a list compr since it's not reused but 🤷 
```suggestion
    await asyncio.gather(
        *[
            api.states.cancel_flow_run(flow_run.id, state_message=msg)
            for flow_run in fr_ids
        ]
    )
```

madkinsz (approved):

src/prefect_server/api/flows.py
cicdw: works for me - will update in a different commit because I forgot to import `asyncio` 🤦

madkinsz (approved):

PR #188, opened 2021-02-16
Upgrade hasura

mashun4ek:
## Summary

Upgrade hasura to 1.3.3.

## Importance

Important to be up to date :)

## Checklist

This PR:

- [-] adds new tests (if appropriate)

### DIFF
--- a/changes/pr188.yaml
+++ b/changes/pr188.yaml
@@ -1,0 +1,2 @@
+enhancement:
+  - "Upgrade hasura version to 1.3.3 - [#188](https://github.com/PrefectHQ/server/pull/188)"
--- a/changes/pr188.yaml
+++ b/changes/pr188.yaml
@@ -1,8 +1,5 @@
 enhancement:
   - "Upgrade hasura version to 1.3.3 - [#188](https://github.com/PrefectHQ/server/pull/188)"
-If you are using managed database (e.g. Cloud SQL, etc) and upgrading hasura image version from 1.3.1 to 1.3.3, make sure that 
-hasura user is the owner of hdb_catalog schema. 
-You can find [more details in hasura documentation](https://hasura.io/docs/1.0/graphql/core/deployment/postgres-requirements.html#notes-for-managed-databases-aws-rds-gcp-cloud-sql-etc).
-Moreover, double check if hasura user owns hdb_catalog.insert_event_log function. 
-To check functions ownership, run following SQL query `SELECT proname, proowner::regrole FROM pg_proc WHERE pronamespace::regnamespace::text = 'hdb_catalog';`.
-To change function ownership, run `ALTER FUNCTION hdb_catalog.insert_event_log OWNER TO hasura;`.
+
+breaking:
+  - "Upgrading hasura version to 1.3.3 for Server that uses managed databases ([readme](https://github.com/PrefectHQ/server/tree/master/helm/prefect-server#Upgrading-Hasura)) - [#188](https://github.com/PrefectHQ/server/pull/188)"
--- a/changes/pr188.yaml
+++ b/changes/pr188.yaml
@@ -1,2 +1,8 @@
 enhancement:
   - "Upgrade hasura version to 1.3.3 - [#188](https://github.com/PrefectHQ/server/pull/188)"
+If you are using managed database (e.g. Cloud SQL, etc) and upgrading hasura image version from 1.3.1 to 1.3.3, make sure that 
+hasura user is the owner of hdb_catalog schema. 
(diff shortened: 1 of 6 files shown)


madkinsz:
Also cleans up some discrepancies between the docker-compose hasura and the helm chart hasura.

madkinsz:
@mashun4ek Would you mind deploying the helm chart with this to make sure it works? We should probably test the prefect CLI too but I'm not sure what the best way to test `prefect server start` is. I think you'd have to build the server image from this commit locally then change the tag in the prefect core code to point to the local image.

mashun4ek:
> @mashun4ek Would you mind deploying the helm chart with this to make sure it works? ~We should probably test the prefect CLI too but I'm not sure what the best way to test `prefect server start` is. I think you'd have to build the server image from this commit locally then change the tag in the prefect core code to point to the local image.~ edit: forgot there was a separate version pin in the prefect repo. the cli can be tested there.
> 
> We'll also need to document / disclaim the changes needed for the Hasura upgrade. The same SQL we ran in Cloud will need to be available at least by reference in the changelog here since people using the helm chart may be using managed databases.

I have tested the helm chart with hasura:1.3.3, it works, no errors or warnings 👍. I added a note about changing ownership for hasura user to changelog, but let me know if I should move it somewhere else.

changes/pr188.yaml
```diff
@@ -0,0 +1,8 @@
+enhancement:
+  - "Upgrade hasura version to 1.3.3 - [#188](https://github.com/PrefectHQ/server/pull/188)"
+If you are using managed database (e.g. Cloud SQL, etc) and upgrading hasura image version from 1.3.1 to 1.3.3, make sure that 
+hasura user is the owner of hdb_catalog schema. 
+You can find [more details in hasura documentation](https://hasura.io/docs/1.0/graphql/core/deployment/postgres-requirements.html#notes-for-managed-databases-aws-rds-gcp-cloud-sql-etc).
+Moreover, double check if hasura user owns hdb_catalog.insert_event_log function. 
+To check functions ownership, run following SQL query `SELECT proname, proowner::regrole FROM pg_proc WHERE pronamespace::regnamespace::text = 'hdb_catalog';`.
+To change function ownership, run `ALTER FUNCTION hdb_catalog.insert_event_log OWNER TO hasura;`.
```
madkinsz, line 8: I'm not sure if this will end up where we want it. I'd put this in the helm chart docs in the "Troubleshooting" section under "Upgrading Hasura"

We may want to link to it from the changelog but I'm not sure.

@cicdw @jlowin interestingly, this is potentially a breaking change for the helm chart but not for other things, should we add a separate breaking change entry that just specifies that this may break a helm chart deployment with an external database?

cicdw: That seems reasonable and more visibility couldn't hurt, although I have a possibly ignorant question -- do people also use the helm chart for upgrading the infrastructure or is it a one-time use for setting it up?

madkinsz: You can run [`helm upgrade`](https://helm.sh/docs/helm/helm_upgrade/) and it's actually kind of smart about just redeploying things that have changed. I use it in development, I imagine some people would use it in their production environments 🤷

changes/pr188.yaml
```diff
@@ -1,8 +1,5 @@
...
   - "Upgrade hasura version to 1.3.3 - [#188](https://github.com/PrefectHQ/server/pull/188)"
-If you are using managed database (e.g. Cloud SQL, etc) and upgrading hasura image version from 1.3.1 to 1.3.3, make sure that 
-hasura user is the owner of hdb_catalog schema. 
-You can find [more details in hasura documentation](https://hasura.io/docs/1.0/graphql/core/deployment/postgres-requirements.html#notes-for-managed-databases-aws-rds-gcp-cloud-sql-etc).
-Moreover, double check if hasura user owns hdb_catalog.insert_event_log function. 
-To check functions ownership, run following SQL query `SELECT proname, proowner::regrole FROM pg_proc WHERE pronamespace::regnamespace::text = 'hdb_catalog';`.
-To change function ownership, run `ALTER FUNCTION hdb_catalog.insert_event_log OWNER TO hasura;`.
+
+breaking:
+  - "Upgrading hasura version to 1.3.3 for Server that uses managed databases ([readme](https://github.com/PrefectHQ/server/tree/master/helm/prefect-server#Upgrading-Hasura)) - [#188](https://github.com/PrefectHQ/server/pull/188)"
```
madkinsz, line 5:
```suggestion
  - "Server instances with managed databases require [manual SQL](https://github.com/PrefectHQ/server/tree/master/helm/prefect-server#Upgrading-Hasura) to upgrade - [#188](https://github.com/PrefectHQ/server/pull/188)"
```

mashun4ek: Sounds much better, thank you!

madkinsz (approved):

madkinsz:
Let's wait for #192 so this can be a separate release

PR #191, opened 2021-02-19
Add tini to graphql and apollo containers

madkinsz:
Finishes #190.

--------

## Summary

Installs `tini` and uses as an entrypoint to graphql/apollo containers to reap zombie processes.

## Notes for review

- Uses sha256 instead of gpg  for checksum verification (both are recommended checks but sha256 is significantly lighter with a 2x faster image build time)

- Fails to run when build on an M1 machine with
```
❯ docker run -it 932c6511928bbb40c364b9a414333e802c9fe8a9edd76dfb7d58cbf59df2c613
/lib64/ld-linux-x86-64.so.2: No such file or directory
```
I've added an `ARCH` flag so you can `docker build . --build-arg ARCH=arm64` for development builds. I presume that once we use apt it'll just grab the correct binary automatically.

## Importance

Fixes PrefectHQ/prefect#4114

## Checklist

This PR:

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -27,7 +27,7 @@ LABEL org.label-schema.version=${PREFECT_SERVER_VERSION}
 LABEL org.label-schema.build-date=${RELEASE_TIMESTAMP}
 
 RUN apt update && \
-    apt install -y gcc git curl && \
+    apt install -y gcc git curl tini && \
     mkdir /root/.prefect/ && \
     pip install --no-cache-dir git+https://github.com/PrefectHQ/prefect.git@${PREFECT_VERSION} && \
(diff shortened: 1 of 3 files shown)


TylerWanner:
Is server currently supported on m1?  wouldn't want to introduce a breaking change here for this.  we can add init to the docker-compose for apollo until we have a multi-arch build ready or we've upgraded the apollo base image to support tini

TylerWanner (approved):

TylerWanner:
this is a fine solution for me

jcrist (approved):

cicdw (approved):

PR #198, opened 2021-02-26
Explicitly use debian buster for apollo/server images

madkinsz:
## Summary

- Upgrades the `apollo` image to use `buster` instead of `stretch` Debian. 
- Changes the `server` image to explicitly use `buster` but should not change anything as it was implicit before

## Importance

- Consistent base OS
- Security patches (see #198) 

## Checklist

This PR:

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,7 @@
 # syntax=docker/dockerfile:1.0.0-experimental
 
 ARG PYTHON_VERSION=${PYTHON_VERSION:-3.7}
-FROM python:${PYTHON_VERSION}-slim
+FROM python:${PYTHON_VERSION}-slim-buster
 
 # Prefect Version, default to MASTER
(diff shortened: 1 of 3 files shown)


services/apollo/healthcheck.js
```diff
@@ -0,0 +1,34 @@
+var http = require("http");
```
cicdw, line 1: jw how is this file related to the image changes?

madkinsz: Whoops! That's not supposed to be in here. I'll remove this :)

madkinsz: I was exploring using node for a healthcheck instead of curl since it's less of a security risk

cicdw: nice - glad i asked haha

cicdw (approved):

PR #199, opened 2021-03-01
Install tini from apt in apollo image

jcrist:
Now that we're using a more recent debian version, we can install `tini`
from apt instead of github.


madkinsz (approved):
Wonderful!

PR #201, opened 2021-03-03
Fix execution of create-tenant-job for custom Apollo ports

jverce:
* Change the `prefect-server.apollo-api-url` template to use the port of
  the Apollo service
* Set the `PREFECT__CLOUD__GRAPHQL` env variable of the
  `create-tenant-job` Job to avoid double concatenation of the
  `/graphql` path when calling the Apollo API

## Summary
Fixed the execution of the `create-tenant-job` in the Helm charts to allow for custom Apollo ports

## Importance
Without this PR, every time a user of the Prefect server Helm chart uses a port for the Apollo service different than the default one (i.e. 4200), and if they set the `jobs.createTenant.enabled` value to `true`, the associated `create-tenant-job` Job will fail with a message similar to this one:
```
requests.exceptions.ConnectTimeout: HTTPConnectionPool(host='prefect-server-apollo.prefect-namespace', port=4200): Max retries exceeded with url: /graphql//graphql (Caused by ConnectTimeoutError(<urllib3.connection.HTTPConnection object at 0x7fb28e69e050>, 'Connection to prefect-server-apollo.prefect-namespace timed out. (connect timeout=15)'))
```

## Checklist

This PR:

### DIFF
--- a/changes/pr200.yaml
+++ b/changes/pr200.yaml
@@ -1,0 +1,7 @@
+fix:
+  - "Fixed the execution of the `create-tenant-job` in the Helm charts to allow
+  for custom Apollo ports -
+  [#200](https://github.com/PrefectHQ/server/pull/200)"
+
+contributor:
+  - "[Jay Vercellone](https://github.com/jverce)"
--- a/changes/pr201.yaml
+++ b/changes/pr201.yaml
@@ -1,7 +1,7 @@
 fix:
   - "Fixed the execution of the `create-tenant-job` in the Helm charts to allow
   for custom Apollo ports -
-  [#200](https://github.com/PrefectHQ/server/pull/200)"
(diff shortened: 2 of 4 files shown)


madkinsz:
Thanks for the PR! 

Hmm... I was under the impression that these ports were _fixed_ because they are the internal. I'll have to take a look at the rest of the code to refresh my understanding of the ports we're setting.

jverce:
In my case, I customised this port because I'm exposing the Apollo service via a load balancer with , and setting the port to 443.
My values file looks something like this:
```yaml
apollo:
  service:
    # type defines the service type and defaults to LoadBalancer
    # to expose the apollo service to users and the UI.
    type: LoadBalancer
    port: 443

  annotations:
    service.beta.kubernetes.io/aws-load-balancer-backend-protocol: http
```

zanieb (approved):

PR #202, opened 2021-03-03
Improve `delete_flow_run` performance by avoiding slow deletion triggers

### LINKED ISSUE #200: Don't rely on cascading deletes when deleting flow runs
## Current behavior
Right now, there are many instances where we delete flow runs; the most important for the purposes of this issue is when toggling a schedule (we delete all auto-scheduled runs in the future), but other situations are archiving, manual deletion, etc.

As users deploy larger and larger Flows, there are situations in which the delete cascades actually timeout (in most circumstances user-facing calls should have low timeout thresholds).  When this occurs, important behavior is not executed that can leave room for confusion and bugs.

## Proposed behavior

We should introduce a delete flow run API function that performs the logic of a cascading delete manually (it should accept a list of flow run IDs); first we delete task run states, then task runs, then the flow run.  Additionally we should audit run deletes and replace them with ID lookups that then are passed to this new API function.

## Example

Reduce timeout scenarios and make deletion more robust.

madkinsz:
## Summary

- Adds a transactional delete to `delete_flow_run` that first deletes `task_run_states` and `task_runs` outer -> innner rather than the default inner -> outer relationships
- Uses the `delete_flow_run` API to delete auto-scheduled flow runs

## Importance

Closes #200 

## Checklist

This PR:

### DIFF
--- a/src/prefect_server/api/runs.py
+++ b/src/prefect_server/api/runs.py
@@ -419,6 +419,15 @@ async def delete_flow_run(flow_run_id: str) -> bool:
     if not flow_run_id:
         raise ValueError("Invalid flow run ID.")
 
+    # Delete task run states and task runs _first_ to speedup deletion which can be
+    # _very_ slow on the large task tables due since they are implemented as
+    # row-triggers and the foreign key checks are slow
+    # TODO: Implement as transaction
+    await models.TaskRunState.where(
+        {"task_run": {"flow_run_id": {"_eq": flow_run_id}}}
+    ).delete()
+    await models.TaskRun.where({"flow_run_id": {"_eq": flow_run_id}}).delete()
+
     result = await models.FlowRun.where(id=flow_run_id).delete()
     return bool(result.affected_rows)  # type: ignore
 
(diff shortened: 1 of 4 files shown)


src/prefect_server/api/runs.py
```diff
@@ -419,8 +421,28 @@ async def delete_flow_run(flow_run_id: str) -> bool:
     if not flow_run_id:
         raise ValueError("Invalid flow run ID.")
 
-    result = await models.FlowRun.where(id=flow_run_id).delete()
-    return bool(result.affected_rows)  # type: ignore
+    # Delete task run states and task runs first to speedup deletion which can be
+    # _very_ slow on the large task tables due since they are implemented as
+    # row-triggers and the foreign key checks are slow
+
+    client = HasuraClient()
```
cicdw, line 428: You'll need to use `prefect.plugins.hasura.client` here so that when this route is called from within Cloud, the right client is used.

cicdw (changes requested):
One-line change request, otherwise LGTM

src/prefect_server/api/runs.py
madkinsz: @cicdw This route isn't called from `Cloud`, it has its own `delete_flow_run` api call

cicdw: Yea, I saw that but would rather not bank on it just in case it gets removed for any reason

cicdw (approved):

PR #207, opened 2021-03-05
Allow agents to be created with empty settings

madkinsz:
Matching GQL API, field is not nullable in the DB

- [ ] Figure out how to use the PR template from the `gh` CLI
- [ ] Add relevant tests
- [x] Add changes entry


cicdw (approved):

cicdw:
@madkinsz FYI creating a changes/ entry when PR'ing against the master branch usually results in the entry being listed as a part of a future release; sorry I merged this quickly before realizing - nbd but just wanted to highlight it

PR #209, opened 2021-03-09
Add inline gh-pages helm repo and CI

madkinsz:
## Summary

Adds a `gh-pages` branch deployed to https://prefecthq.github.io/server/ with an `index.yaml` and `charts/prefect-server-<version>.tgz` to create a working helm repository! On tagged pushes, a new helm chart version will be pushed to the `gh-pages` branch via Github Action. See example build at https://github.com/PrefectHQ/server/runs/2078623350

Allows use of the helm chart with `helm repo add prefecthq https://prefecthq.github.io/server/ ` and `helm install prefecthq/prefect-server`

See working readme at https://github.com/PrefectHQ/server/blob/helm-repo/helm/prefect-server/README.md

This also updates the helm chart slightly to allow for easier versioning using `appVersion` and creates the `serverVersionTag` for install-time overrides and distinguishes `prefectVersionTag` for pinning versions for items that use core image tags.

After merge, I will delete the test tarballs from the helm repository as they use tags (eg `0.1.0`) that do not exist. For testing, you'll have to include `--set serverVersionTag=2021.03.06` in your installation command.

## Importance

- Makes the helm chart a little easier to install
- Adds versioning to the helm chart
- The helm chart is official now!

## Future work

Prettyify the gh-pages website similar to `dask-gateway`

## Checklist

This PR:


helm/prefect-server/README.md
```diff
@@ -1,31 +1,132 @@
...
+    ```
+
+    _A name of your choice can be provided instead of `--generate-name`_
+
+    _If chart installation fails, `--debug` can provide more information_
+
+
+#### Verifying package integrity
+
+The hosted charts are signed using GPG to ensure package integrity.
```
jcrist, line 68: I might mark this as optional, gpg stuff could scare off new users.

helm/prefect-server/README.md
```diff
@@ -1,31 +1,132 @@
...
+3. Search for available charts and versions
+
+    ```
+    $ helm search repo [name]
+    ```
+
+4. Install the Helm chart
+
+    ```
+    $ helm install prefecthq/prefect-server --generate-name
```
jcrist, line 58: I might add a few more common options to the example install command (including the release name, values file, and namespace). You might also encourage the use of `helm upgrade --install` instead of `helm install`, as the former can be used for both installs and upgrades. 

See e.g. https://gateway.dask.org/install-kube.html#install-the-helm-chart for what I'd expect:

```
helm upgrade --install \
    --namespace $NAMESPACE \
    --version $VERSION \
    --values path/to/your/config.yaml \
    $RELEASE \
    prefecthq/prefect-server
```

madkinsz: Thanks for taking a look. Those docs look a lot more approachable than this page but I don't think I'll go much further for now. I've taken a shot at this request though and I agree it helps.

cicdw (approved):

PR #211, opened 2021-03-10
Fix Helm UI and Core version tags

madkinsz:
## Summary

Of course, in #209 I was testing a tag that _also_ existed in the UI but the UI is not guaranteed to be released at the same time as Server. Therefore the UI needs to rely on a different tag.

Now, we'll pull the latest core and UI tags and set them in `values.yaml` for the agent/ui.

## Importance

The existing helm chart is broken :( will need to be re-released.

## Checklist

This PR:

- ~[ ] adds a change file in the `changes/` directory (if appropriate)~

### DIFF
--- a/helm/prefect-server/templates/agent/deployment.yaml
+++ b/helm/prefect-server/templates/agent/deployment.yaml
@@ -36,8 +36,7 @@ spec:
         securityContext:
           {{- toYaml . | nindent 8 }}
         {{- end }}
-        {{- /* We need to trim `core-` off the prefectVersionTag to include the K8s orchestration extras */}}
-        image: "{{ .Values.agent.image.name }}:{{ .Values.agent.image.tag |  default (.Values.prefectVersionTag | replace "core-" "") }}"
+        image: "{{ .Values.agent.image.name }}:{{ .Values.agent.image.tag | default .Values.prefectVersionTag }}"
         imagePullPolicy: {{ .Values.agent.image.pullPolicy  }}
         command:
           - bash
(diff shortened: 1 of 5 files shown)


cicdw (approved):

helm/prefect-server/templates/agent/deployment.yaml
```diff
@@ -36,7 +36,8 @@ spec:
         securityContext:
           {{- toYaml . | nindent 8 }}
         {{- end }}
-        image: "{{ .Values.agent.image.name }}:{{ .Values.agent.image.tag |  default .Values.prefectVersionTag }}"
+        {{- /* We need to trim `core-` off the prefectVersionTag to include the K8s orchestration extras */}}
+        image: "{{ .Values.agent.image.name }}:{{ .Values.agent.image.tag |  default (.Values.prefectVersionTag | replace "core-" "") }}"
```
jcrist, line 40: Why not set the default image name in the chart values  to the latest core release (instead of `null`) when releasing the helm chart (since we already pull the core version as part of the release process)? May be cleaner.

madkinsz: I can't think of a simple way to target the agent image with `sed`. Perhaps worth revisiting if we add a Python build-tool into the mix that can just load the yaml.

madkinsz: Actually now that we're not using `core-X` to pull a UI version we can just set the `prefectVersionTag` without `core`

madkinsz: Although I may want to strip the `core-` here anyway? I'm not sure what's best here

jcrist: If a user provides something that's invalid, we should error, not implicitly assume we know what they want. What if it's a non-prefect image that starts with `core-`?

PR #215, opened 2021-03-15
Update scheduler to have deterministic order

jlowin:
## Summary

The scheduler currently orders flows by maximum scheduled start time, such that a flow (A) with its last scheduled run starting in one hour is prioritized over a flow (B) with its last scheduled run starting in one day.

However, this order could lead to flows being skipped across multiple batches. Consider if `A` is scheduled in the first batch of 500 flows, and `B` is the 501st flow. The scheduler schedules `A` such that it's maximum scheduled start time becomes later. This makes `B` the 500th flow, not the 501st. Therefore, when the scheduler queries for the second batch, it doesn't pick up `B`.

This circumstance is highly unusual due to the practical manner in which scheduling works (flows don't typically get to a point where they have a very low maximum scheduled start time and then suddenly a very high maximum scheduled start time), but it is possible and this PR closes it with a deterministic sort by ID. In order to reduce any minor concern about prioritizing flows in need of scheduling, it also doubles the scheduler loop interval.

## Checklist

This PR:

### DIFF
--- a/changes/pr215.yaml
+++ b/changes/pr215.yaml
@@ -1,0 +1,20 @@
+# An example changelog entry
+#
+# 1. Choose one (or more if a PR encompasses multiple changes) of the following headers:
+#   - feature
+#   - enhancement
+#   - fix
+#   - deprecation
+#   - breaking (for breaking changes)
+#   - migration (for database migrations)
+#
+# 2. Fill in one (or more) bullet points under the heading, describing the change.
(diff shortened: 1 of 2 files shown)


madkinsz (approved):
>  it also doubles the scheduler loop interval.

halves the interval, doubles the amount of runs

lgtm!

jlowin:
@madkinsz you're right 🤦

Issue #216, opened 2021-03-15
Cannot set Prefect Server logging level


peterroelants:
## Description
Setting Prefect Server logging level via environment variable doesn't seem to work. Log level stays at "INFO".

Neither `PREFECT__LOGGING__LEVEL: WARNING` nor `PREFECT_SERVER__LOGGING__LEVEL: WARNING` seemed to have the desired effect.

## Expected Behavior
I expect the Prefect Server to follow the [Prefect Logging Configuration documentation](https://docs.prefect.io/core/concepts/logging.html).

## Reproduction
Add `PREFECT__LOGGING__LEVEL: WARNING` or `PREFECT_SERVER__LOGGING__LEVEL: WARNING` to the `graphql` `environment` and start the prefect server.

Note the "INFO" logging messages, e.g.:
```
...
graphql_1   | INFO:     192.168.32.4:40048 - "GET /health HTTP/1.1" 200 OK
graphql_1   | INFO:     192.168.32.4:40052 - "GET /health HTTP/1.1" 200 OK
...
```

## Environment
Server docker image: `prefecthq/server:core-0.14.12` with ID `f92fdce05361`

peterroelants:
Issue is related to uvicorn not being launched with logging level from prefect config.

I've created a fix at https://github.com/PrefectHQ/server/pull/217

PR #217, opened 2021-03-15
Set uvicorn server log level

peterroelants:
## Summary
Set graphql uvicorn server log level based on Prefect config.

Solves https://github.com/PrefectHQ/server/issues/216

## Importance
Allows the user to control the access logs of the Prefect GraphQL service (which can become quite large on the default `DEBUG` level).

## Checklist

This PR:

- [x] adds new tests (N/A)

### DIFF
--- a/src/prefect_server/services/graphql/server.py
+++ b/src/prefect_server/services/graphql/server.py
@@ -51,9 +51,24 @@ def health(request: Request) -> JSONResponse:
     return JSONResponse(dict(status="ok", version=app_version))
 
 
+def _get_uvicorn_log_level():
+    prefect_log_level = prefect_server.configuration.config["logging"]["level"]
+    # Uvicorn log levels are lower case
+    uvicorn_log_level = prefect_log_level.lower()
+    if uvicorn_log_level in uvicorn.config.LOG_LEVELS:
+        logger.info(f"Using uvicorn log level = {repr(uvicorn_log_level)}")
+        return uvicorn_log_level
+    else:
+        logger.warning(
+            f"{repr(uvicorn_log_level)} not a valid uvicorn log level, falling back on default."
+        )
+        return None
+
+
 if __name__ == "__main__":
     uvicorn.run(
         app,
         host=prefect_server.config.services.graphql.host,
         port=prefect_server.config.services.graphql.port,
+        log_level=_get_uvicorn_log_level(),
     )
--- a/src/prefect_server/services/graphql/server.py
+++ b/src/prefect_server/services/graphql/server.py
@@ -60,7 +60,7 @@ def _get_uvicorn_log_level():
         return uvicorn_log_level
     else:
         logger.warning(
-            f"{repr(uvicorn_log_level)} not a valid uvicorn log level, falling back on default."
+            f"{uvicorn_log_level!r} not a valid uvicorn log level, falling back on default."
         )
         return None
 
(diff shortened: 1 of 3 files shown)


src/prefect_server/services/graphql/server.py
```diff
@@ -51,9 +51,24 @@ def health(request: Request) -> JSONResponse:
     return JSONResponse(dict(status="ok", version=app_version))
 
 
+def _get_uvicorn_log_level():
+    prefect_log_level = prefect_server.configuration.config["logging"]["level"]
+    # Uvicorn log levels are lower case
+    uvicorn_log_level = prefect_log_level.lower()
+    if uvicorn_log_level in uvicorn.config.LOG_LEVELS:
+        logger.info(f"Using uvicorn log level = {repr(uvicorn_log_level)}")
```
madkinsz, line 59: You can do quick reprs in Python string formatting
```suggestion
        logger.info(f"Using uvicorn log level = {uvicorn_log_level!r}")
```

src/prefect_server/services/graphql/server.py
```diff
@@ -51,9 +51,24 @@ def health(request: Request) -> JSONResponse:
...
+        return uvicorn_log_level
+    else:
+        logger.warning(
+            f"{repr(uvicorn_log_level)} not a valid uvicorn log level, falling back on default."
```
madkinsz, line 63:
```suggestion
            f"{uvicorn_log_level!r} not a valid uvicorn log level, falling back on default."
```

madkinsz (commented):
This seems like a really reasonable fix to me, just some minor nits.

I think it is worth noting that:
- We can disable access logs to Uvicorn without changing the log level, this might be good to add to the config as `server.logging.log_access = false` and pass to Uvicorn
- We probably will increase the healthcheck interval because it's pretty fast

Thanks so much for contributing!

peterroelants:
@madkinsz Thank you for the quick review and the `repr` f-string suggestion, I didn't know about that.

I added a `server.logging.log_access` logging level as you suggested. However, I'm unsure if it should be at `logging.log_access`, or if it would be better to provide a `graphql.logging.log_access` config field. Do you have any opinion on this?

madkinsz:
> @madkinsz Thank you for the quick review and the `repr` f-string suggestion, I didn't know about that.

You're welcome :) Also, I left this as is before because I didn't think it was a big deal, but when using the config it's nice to use the attribute (`.`) access pattern e.g. `prefect_server.configuration.config.server.logging.level` instead of item access. They're equivalent but staying consistent with the rest of the repo makes it easier to search for usage of config values in the code.

> I added a `server.logging.log_access` logging level as you suggested. However, I'm unsure if it should be at `logging.log_access`, or if it would be better to provide a `graphql.logging.log_access` config field. Do you have any opinion on this?

Great point! This would fit better under `[services.graphql]`. Perhaps we should name it `services.graphql.disable_access_logs = True` to match the Uvicorn name.

> In the case of `graphql.logging.log_access` we might also want to change the logging level change to influence `graphql.logging.level`?

I think the rename suggested above will make it clear that you're just disabling it regardless of log level (as in Uvicorn). I don't think we should mutate another config value based on it.

peterroelants:
> when using the config it's nice to use the attribute (`.`) access pattern e.g. `prefect_server.configuration.config.server.logging.level` instead of item access. They're equivalent but staying consistent with the rest of the repo makes it easier to search for usage of config values in the code.

I see now you are using box. I updated the access pattern as you suggested.

> Great point! This would fit better under `[services.graphql]`. Perhaps we should name it `services.graphql.disable_access_logs = True` to match the Uvicorn name.

Good suggestion, I updated the code with this.

madkinsz (approved):
Thanks again!

PR #219, opened 2021-03-17
Enforce idempotency via database constraint

jlowin:
## Summary

Idempotency is currently enforced in the application but this is prone to race conditions under certain circumstances. This PR moves the logic to the database itself and cleanly applies it via upsert in `create_flow_run()`. 

Depends on #218 

## Checklist

This PR:


zanieb (commented):
I'm a bit confused about the case where the idempotency key is `None`

jlowin:
`NULL` doesn't violate a unique constraint, so if the key is `None` a normal insert takes place.

src/prefect_server/api/runs.py
```diff
@@ -227,26 +174,39 @@ async def _create_flow_run(
...
         scheduled_start_time=scheduled_start_time,
         name=flow_run_name or names.generate_slug(2),
-        states=[
-            models.FlowRunState(
-                tenant_id=flow.tenant_id,
-                **models.FlowRunState.fields_from_state(
-                    Pending(message="Flow run created")
-                ),
-            )
-        ],
```
madkinsz, line 237: No more pending state?

src/prefect_server/api/runs.py
```diff
@@ -287,9 +247,10 @@ async def get_or_create_task_run(
             states=[
                 models.TaskRunState(
                     tenant_id=task.tenant_id,
-                    **models.TaskRunState.fields_from_state(
-                        Pending(message="Task run created")
-                    ),
+                    state="Pending",
+                    timestamp=pendulum.now(),
+                    message="Task run created",
+                    serialized_state=Pending(message="Task run created").serialize(),
```
madkinsz, line 253: Are we moving away from this pattern?

madkinsz: (I see you've deleted it below -- alrighty)

zanieb (commented):
A couple minor questions

src/prefect_server/api/runs.py
jlowin: Yeah, it's an unnecessary function that hides what's actually happening.

src/prefect_server/api/runs.py
jlowin: We can't do the nested insert here because it will succeed even if the idempotency key is hit (due to the ON CONFLICT clause), so it's now done [here](https://github.com/PrefectHQ/server/pull/219/files#diff-aaf8abba6b80fc268ac079173dda255a41f8bd6ee121c9101229771497d63321R196). Since all flow runs (at this time!) go immediately into a scheduled state, there's no need for the intermediate pending one. In the future if we want to create unscheduled flow runs (for your backend `flow.run()` for example) then we can toggle out the `Scheduled` state for a `Pending` one.

zanieb (approved):

PR #225, opened 2021-03-19
Increase number of max scheduled runs to 50

znicholasbrown:
## Summary

Increases the max number of scheduled runs from 10 -> 50

## Importance

Gives the scheduler some breathing room when running over large numbers of flows.

## Checklist

This PR:

- [cc @cicdw] adds new tests (if appropriate)

### DIFF
--- a/src/prefect_server/api/flows.py
+++ b/src/prefect_server/api/flows.py
@@ -556,7 +556,7 @@ async def schedule_flow_runs(flow_id: str, max_runs: int = None) -> List[str]:
     """
 
     if max_runs is None:
-        max_runs = 10
+        max_runs = 50
 
     if flow_id is None:
         raise ValueError("Invalid flow id.")


bouke-sf:
#281 makes this value configurable.

madkinsz:
Note that this was left open because raising it to 50 introduced some other issues, you may want to avoid changing the value to be too large @bouke-sf

bouke-sf:
Thanks for mentioning! Were the issues specifically caused by 1 flow with 50 runs, or just 50 concurrent runs in general?

madkinsz:
I don't remember, actually :) cc @znicholasbrown if you remember what issues arose

znicholasbrown:
@bouke-sf - the primary issue is just related to scheduler performance; the scheduler will need to loop through every flow in the db and can back up if it's not able to complete that loop before it's called again; this should only impact pretty large deployments though. Love the configurability!

PR #235, opened 2021-03-31
Update `LoopServices` to use the config to specify loop times

madkinsz:
## Summary

This appears to be a necessary first step for tightening loop service times to reduce service test failures in CI. Once we're using the config instead of 'loop_seconds_default' then we can set a config for CI that uses faster loop times. This will require changes in Cloud as well.

- Adds `get_backend_config()` plugin so we can load the config across Server/Cloud
- Adds default config key inference and precedence to `LoopService` config key loading

## Importance

Improve test times and reduce timeout failures in Cloud https://github.com/PrefectHQ/cloud/issues/3065

## Checklist

This PR:

### DIFF
--- a/changes/pr235.yaml
+++ b/changes/pr235.yaml
@@ -1,0 +1,3 @@
+
+enhancement:
+  - "Update `LoopServices` to use the config to specify loop times - [#235](https://github.com/PrefectHQ/server/pull/235)"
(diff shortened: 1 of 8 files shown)


madkinsz:
Closing in favor of a simpler implementation at #242

PR #238, opened 2021-04-05
Break up task edge registration

zangell44:
## Summary
This PR batches `Task` and `Edge` creation when creating a `Flow`.

## Importance
For very large flows, we're hitting a limit on Postgres query size when trying to create the full Flow model. See this Slack thread for a full discussion https://prefect-community.slack.com/archives/C014Z8DPDSR/p1617128789186300.

I think eventually we'll want to do this batching on the client side, but this PR along with [an update to speed up flow validation](https://github.com/PrefectHQ/prefect/pull/4347) will allow registration of fairly large flows for the time being.

## Checklist

This PR:


src/prefect_server/api/flows.py
```diff
@@ -261,37 +266,56 @@ async def create_flow(
...
+                    for e in edges_chunk
+                ]
             )
-            for e in flow.edges
-        ],
-    ).insert()
+
+    except Exception as exception:
+        await delete_flow(flow_id=flow_id)
+        raise exception
```
madkinsz, line 318: We might want to reraise this as an `APIError` although I'm not sure where we want to do this generally. Perhaps it belongs in the graphql layer. cc @cicdw

src/prefect_server/utilities/utils.py
```diff
@@ -0,0 +1,22 @@
+from typing import Iterable
+
+
+def get_chunk(iterable: Iterable, chunk_size: int):
```
madkinsz, line 4: I think this would be better in a `collections.py` file. I think there may be some pushback against creating a new file for something that's only used once but 🤷

madkinsz: I think I like the implementation at https://alexwlchan.net/2018/12/iterating-in-fixed-size-chunks/ more. It feels weird to build an intermediate list then cast it to a tuple. Another note that I find that name better as well as `get_chunk` makes me feel like you're returning a single chunk, I get that it's yielding but this isn't how std python names generator functions.

madkinsz: We also have a dependency on `toolz` which has https://toolz.readthedocs.io/en/latest/api.html#toolz.itertoolz.partition which looks like it may be slightly less performant.

tests/api/test_flows.py
```diff
@@ -514,6 +516,36 @@ async def test_create_flow_persists_serialized_flow(self, project_id, flow):
...
 
+    async def test_create_flow_cleans_up_if_task_creation_fails(
+        self, project_id, flow, monkeypatch
+    ):
+        patched_insert_raises_error = MagicMock(side_effect=Exception())
+        monkeypatch.setattr(
+            "prefect.models.Task.insert_many", patched_insert_raises_error
+        )
+        patched_delete = MagicMock()
+        monkeypatch.setattr("prefect_server.api.flows.delete_flow", patched_delete)
```
madkinsz, line 527: I think we should assert that it's deleted rather than using a mock.

tests/api/test_flows.py
```diff
@@ -514,6 +516,36 @@ async def test_create_flow_persists_serialized_flow(self, project_id, flow):
...
+        with pytest.raises(Exception):
+            flow_id = await api.flows.create_flow(
+                project_id=project_id, serialized_flow=flow.serialize()
+            )
+            assert patched_delete.called
+
+    async def test_create_flow_cleans_up_if_edge_creation_fails(
```
madkinsz, line 534: These seems like a good case for a parametrized test on `"Edge"` and `"Task"` but I do not feel strongly about it.

changes/pr238.yaml
```diff
@@ -0,0 +1,21 @@
...
+# 2. Fill in one (or more) bullet points under the heading, describing the change.
+#    Markdown syntax may be used.
+#
+# 3. If you would like to be credited as helping with this release, add a
+#    contributor section with your name and github username.
+#
+# Here's an example of a PR that adds an enhancement
+
+enhancement:
+  - "Batch task and edge creation - [#238](https://github.com/PrefectHQ/server/pull/238)"
```
madkinsz, line 20:
```suggestion
  - "Batch task and edge insertion during flow creation - [#238](https://github.com/PrefectHQ/server/pull/238)"
```

madkinsz: I would _consider_ adding a `fix` item as well i.e. "Allow flows with >1500 tasks to register without timeout"

madkinsz (changes requested):
Good sleuthing!

src/prefect_server/api/flows.py
```diff
@@ -261,37 +266,56 @@ async def create_flow(
...
+                    )
+                    for e in edges_chunk
+                ]
             )
-            for e in flow.edges
-        ],
-    ).insert()
+
+    except Exception as exception:
+        await delete_flow(flow_id=flow_id)
```
cicdw, line 317: This needs to be `prefect.api.flows.delete_flow` so that Cloud can override the logic here

src/prefect_server/api/flows.py
cicdw: Hmmmm yea, at a minimum I'd vote to log this error and raise a custom exception "Failed to register Flow" instead of raising the generic exception -- an APIError would work too

tests/api/test_flows.py
cicdw: Agreed -- one way to do this would be to lookup the name of this flow in the DB and confirm that it is not present before and after your `create_flow` call.

cicdw (changes requested):
Great work digging into the subtleties here both client and server side!

src/prefect_server/utilities/utils.py
zangell44: All fair points thanks, I went with your preferred implementation in the latest commit

tests/api/test_flows.py
zangell44: might as well use this opportunity to get more familiar with pytest parameterization!

tests/api/test_flows.py
zangell44: went with Chris' suggestion in the latest commit

zangell44:
All feedback has been addressed, albeit likely imperfectly

Would be good to get another set of eyes on the exception handling here https://github.com/PrefectHQ/server/pull/238/files#diff-23db4a75ea24edcd454c5f66ca3e1bf042862bc3713786a8924acd5e06cc256dR320. Slightly confused on what the best pattern for this would be, but logging the error and raising a custom exception from the original seemed somewhat reasonable

src/prefect_server/api/flows.py
```diff
@@ -261,37 +267,57 @@ async def create_flow(
...
+                ]
             )
-            for e in flow.edges
-        ],
-    ).insert()
+
+    except Exception as exception:
+        logger.error(exception)
+        await api.flows.delete_flow(flow_id=flow_id)
+        raise Exception("Failed to register Flow") from exception
```
cicdw, line 320: @madkinsz might have additional thoughts but we should raise some specific error here (probably an `APIError`) instead of a generic exception

tests/api/test_flows.py
```diff
@@ -514,6 +516,22 @@ async def test_create_flow_persists_serialized_flow(self, project_id, flow):
...
 
+    @pytest.mark.parametrize("model_name", ["Task", "Edge"])
+    async def test_create_flow_cleans_up_if_task_or_edge_creation_fails(
+        self, project_id, flow, monkeypatch, model_name
+    ):
+        patched_insert_raises_error = MagicMock(side_effect=Exception())
+        monkeypatch.setattr(
+            f"prefect.models.{model_name}.insert_many", patched_insert_raises_error
+        )
+        flow.name = "my special flow"
```
cicdw, line 527: jw - did you change the name on the flow to ensure the name was unique?

cicdw (changes requested):
Some really minor comments, overall this is looking good to me

src/prefect_server/api/flows.py
madkinsz, line 320: I would do this

```suggestion
    except Exception as exc:
        logger.error("`create_flow` failed during insertion", exc_info=True)
        await api.flows.delete_flow(flow_id=flow_id)
        raise APIError() from exc
```

- The log now explains what happens and attaches a full traceback
- The `APIError` will hide all further exception info from the user but still be chained to the original exception for tracking in GCP

tests/utilities/test_collections.py
```diff
@@ -0,0 +1,12 @@
+from prefect_server.utilities.collections import chunked_iterable
+
+
+def test_get_chunk_of_list():
```
madkinsz, line 4: Should probably rename these to match

tests/api/test_flows.py
zangell44: yes, I wasn't sure what precisely our pytest logic for setup/teardown during tests was

src/prefect_server/api/flows.py
zangell44: love it

madkinsz (approved):

PR #242, opened 2021-04-12
Simplify `LoopService.loop_seconds` handling and add debug mode to speedup tests

madkinsz:
## Summary

- Removes config key `loop_seconds` loading
- Removes `__init__``loop_seconds` passing 
- Adds getter/setter for `loop_seconds`
- Adds `loop_seconds_debug` that is set during tests by default

## Checklist

This PR:

### DIFF
--- a/src/prefect_server/services/loop_service.py
+++ b/src/prefect_server/services/loop_service.py
@@ -18,7 +18,7 @@ class LoopService:
     """
 
     # The number of seconds the loop should repeat on when in operation
-    loop_seconds = 600
+    loop_seconds_default = 600
 
     # The number of seconds the loop should repeat on when testing
     loop_seconds_debug = 1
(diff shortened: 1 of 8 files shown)


src/prefect_server/services/loop_service.py
```diff
@@ -16,35 +17,35 @@ class LoopService:
...
+            self.loop_seconds_debug
+            if os.environ.get(self.debug_environment_key) is not None
+            else self._loop_seconds
+        )
+
+    @loop_seconds.setter
+    def loop_seconds(self, value):
+        if value == 0:
+            raise ValueError("`loop_seconds` must be greater than 0.")
+        self._loop_seconds = float(value)
```
madkinsz, line 47: I noticed some code in the Cloud batch writer that also implements these properties which means retaining `_default` was the easiest way to avoid naming collisions. I'll probably rewrite those to call `super().` in their default cases.

src/prefect_server/services/loop_service.py
```diff
@@ -16,35 +17,35 @@ class LoopService:
...
-            raise ValueError("`loop_seconds` must be greater than 0.")
-
-        self.loop_seconds = float(loop_seconds)
+    def __init__(self):
+        self.loop_seconds = self.loop_seconds_default
         self.name = type(self).__name__
         self.logger = utilities.logging.get_logger(self.name)
 
+    @property
+    def loop_seconds(self):
```
madkinsz, line 36: I noticed some code in the Cloud batch writer that also implements this property which made me feel retaining the `_default` suffix on `loop_seconds` was the easiest way to avoid naming collisions. I'll probably rewrite those to call `super().` in their default cases.

src/prefect_server/services/loop_service.py
```diff
@@ -16,35 +17,35 @@ class LoopService:
     define the `run_once` coroutine to describe the behavior of the service on each loop.
     """
 
-    # if set, and no `loop_seconds` is provided, the service will attempt to load
-    # `loop_seconds` from this config key
-    loop_seconds_config_key = None
```
madkinsz, line 21: This feature is unused and ought to be removed for simplification

src/prefect_server/services/loop_service.py
```diff
@@ -16,35 +17,35 @@ class LoopService:
...
     loop_seconds_default = 600
 
+    # The number of seconds the loop should repeat on when testing
+    loop_seconds_debug = 1
+    debug_environment_key = "PYTEST_CURRENT_TEST"
+
     # shutdown flag for gracefully exiting the infinite loop
     is_running = True
 
-    def __init__(self, loop_seconds: Union[float, int] = None):
```
madkinsz, line 29: Passing loop seconds to `__init__` is unused and is being removed for simplification.

jlowin (approved):

madkinsz:
@jlowin I realized that we start services separately from tests in the integration/service test CI steps so I've switched to a different environment variable than the pytest one.

jlowin (approved):

PR #268, opened 2021-08-02
Fix Apollo URL environment variable

nebur1989:
## Summary
According to the TOML standard, this environment variable was lacking the UI prefix.
Making the UI not properly working.

## Importance
This pull request aims to fix the UI, that is currently not working.

## Checklist
This PR:
- [X] adds new tests (if appropriate)

### DIFF
--- a/changes/pr268.yaml
+++ b/changes/pr268.yaml
@@ -1,0 +1,5 @@
+fix:
+  - "Fix Apollo URL environment variable - [#268](https://github.com/PrefectHQ/server/pull/268)"
+
+contributor:
+  - "[Ruben Aguilar](https://github.com/nebur1989)"
--- a/helm/prefect-server/templates/ui/deployment.yaml
+++ b/helm/prefect-server/templates/ui/deployment.yaml
@@ -40,7 +40,7 @@ spec:
           command: 
             - "/intercept.sh"
           env:
-            - name: PREFECT_SERVER__APOLLO_URL
+            - name: PREFECT_SERVER__UI__APOLLO_URL
               value: {{ .Values.ui.apolloApiUrl }}
           ports:
             - name: http


madkinsz (commented):
Hey @nebur1989 -- this is actually correct as-is as we're setting the Apollo URL in the Prefect UI container. See https://github.com/PrefectHQ/ui/blob/master/start_server.sh

afranzi:
Hi @madkinsz, the documentation in https://docs.prefect.io/orchestration/server/deploy-local.html#ui-configuration mentions that the property is being used inside the `server --> server.ui` section. To replace this property properly and following the toml standards you have to overwritten by using the `PREFECT_SERVER__UI__APOLLO_URL` instead of `PREFECT_SERVER__APOLLO_URL`.

```
[server]
  [server.ui]
  apollo_url="http://localhost:4200/graphql"
```

madkinsz:
@afranzi That configuration is for the prefect CLI to spin up a server instance which is quite distinct from the helm chart. You can even see when the CLI sets this value in the `docker-compose` it names it the same as here. https://github.com/PrefectHQ/prefect/blob/master/src/prefect/cli/docker-compose.yml#L130 https://github.com/PrefectHQ/prefect/blob/master/src/prefect/cli/server.py#L306

This is definitely as intended and functioning correctly as-is.

I appreciate the desire to contribute; hope to see you around here still.

afranzi:
The docker-compose file is wrong too 😬 . I believe it should use also the env var with the `__UI__` part.

The server-client file is extracting the configuration from https://github.com/PrefectHQ/prefect/blob/master/src/prefect/config.toml#L48 which contains the property in the `server.ui` section.

Are you sure the docker-compose is working as expected?

afranzi:
I was checking the docker-compose file and the config.toml and it not consistent at all.

https://github.com/PrefectHQ/prefect/blob/master/src/prefect/cli/docker-compose.yml#L66-L68

There are some properties (env vars) using the right format, so they can overwrite the [server.database] and [server.hasura] properties.
```
      PREFECT_SERVER__DATABASE__CONNECTION_URL: ${DB_CONNECTION_URL}
      PREFECT_SERVER__HASURA__ADMIN_SECRET: ${PREFECT_SERVER__HASURA__ADMIN_SECRET:-hasura-secret-admin-secret}
      PREFECT_SERVER__HASURA__HOST: hasura
```
But the `[server.ui]` is not defined properly.

madkinsz:
The `prefect` `config.toml` has Server and UI sections to set environment variables for the UI and Server containers. The UI and Server containers do not know anything about the `config.toml` in the `prefect` repo and do not care about its hierarchy. Prefect Server defines its own `config.toml` that it uses for settings (see `PREFECT_SERVER` vs `PREFECT__SERVER`) and the UI uses the environment variable in `start_server.sh` to set the apollo connection url. These are definitely set up correctly and the UI issue was unrelated and should be resolved now.

PR #281, opened 2021-08-31
Add max runs per flow parameter

### LINKED ISSUE #280: Make max number of scheduled runs per flow configurable
## Opened from the [Prefect Public Slack Community](https://prefect.io/slack)

**bouke.krom**: Good morning! We are trying to run a single flow with many (about 30) different parameters at the same schedule: all Monday morning 9am. This morning exactly 10 runs were performed, the others were silently skipped. We run Prefect server with a single LocalAgent. It seems like we're running into some default 'max 10' setting. The agent displays as having 10 'submittable runs' -- maybe this can be increased but no hits for `submittable` in the docs... Any pointers?

**wilson.bilkovich**: Looking at the code, there does seem to be a limit, but it looks to be `100` not 10.. I don’t see the text ‘submittable’ anywhere in the client or server repos; can you paste an exact string?

**bouke.krom**: 

**bouke.krom**: But then again this flow (the only one to be executed by this agent) claims only 10 upcoming runs too:

**bouke.krom**: 

**wilson.bilkovich**: 
```
query AgentSubmittableRuns {
    flow_run (where: {state: {_eq: "Scheduled"}}) {
    id
    name
    state
    labels
    scheduled_start_time
    version
    flow_id
    }
}
```

…looks to be the query it runs to get the submittable jobs, which doesn’t impose a limit

**wilson.bilkovich**: What happens when you click on the `Late` icon? Are the rest of them shown there by any chance?

**bouke.krom**: Hey thanks for the help! No there are no `late` flows

**wilson.bilkovich**: Weird; I don’t see anything under the hood that should be limiting you.  Hopefully someone more knowledgable has an idea.

**kevin701**: Hey <@U028VSXD87Q>, are you on server ot cloud?

**bouke.krom**: server

**kevin701**: What version are you on?

**bouke.krom**: 0.15.0

**kevin701**: I can’t see why there would be any limit. When you say they were skipped, did the flow runs just disappear or did they fail? Is your local agent also capable of handling more than 10 concurrent flows?

**bouke.krom**: Found it! The limit of max 10 into the future per flow is defined here: <https://github.com/PrefectHQ/server/blob/93b08f97558ac3c620e5ab9e44b781daf23eb9dc/src/prefect_server/api/flows.py#L598>
• For our use case I would like to increase the number (at least to 100)
• For general use, I'd say that this is a bug. Scheduling more than 10 runs of the same flow at the same moment will fail silently. The scheduling happens here: <https://github.com/PrefectHQ/server/blob/93b08f97558ac3c620e5ab9e44b781daf23eb9dc/src/prefect_server/services/towel/scheduler.py#L20> It queries 500 flows at a time so the limit of 10 per flow seems out of balance
    ◦ Increase 10 to 100?
    ◦ Even then make sure that flow runs that are skipped are noticed 
I'd be happy to contribute, but this probably involves a bit more than just PR'ing the hardcoded `10` . What is the right course of action <@U01QEJ9PP53> <@UKTUC906M>?

**kevin701**: Will ask the team about this

**kevin701**: Great digging btw! Would you be willing to make a PR? It will probably look like
• creating a new setting in `config.toml` (in the server repo), something like `max_runs_to_schedule_per_flow`
• replacing the hard coded 10 with this setting

**kevin701**: The default of 10 would still be good  (some people take down their infrastructure)

**bouke.krom**: Thanks! Yes I can make a PR -- I would need to read up on how the `config.toml` is used etc.
Just to check though:
• To me it would make more sense if there is a 'time to schedule into the future' (e.g. 1 week ahead or something). This complicates the fix but it could remove the difference in scheduling between 1 or several flows. Is there any appetite for that?
&gt;  some people take down their infrastructure
Not sure what you mean. Would there be a downside to increasing the default of 10? Since I'm the first one running into this for 13 months it probably doesnt make a difference to anyone :thinking_face:

**kevin701**: Isn’t it the other way around? You’re the first person to run into it (in the last couple of months) so we don’t need to increase the default? :sweat_smile:

Not 100% sure what you mean about scheduling int he future? We do have `start_dates` attached to schedules

**kevin701**: And what I meant was that people in the past have run into issues where running multiple concurrent flows kills their machines.

**bouke.krom**: Yes, the inverse is also true :smile:

**bouke.krom**: Currently the code schedules the first 10 runs for every flow. This only protects against concurrent flow run problems if the runs are from the same flow. The other reason for limiting the number of flows is that schedules can plan ahead indefinitely, so you'd get endless lists. Introducing a parameter with a default is yet another way things can go wrong. Another approach could be to keep scheduling runs for every flow until they are too far into the future, e.g. plan every flow run for the coming week. This 'planning horizon' can be a setting but it only matters for the UI display. That way you're sure every run will be scheduled and there's no setting to get wrong.

**zach555**: <@U028VSXD87Q> for consistency, I would start with adding a default in `config.toml` and replacing the hard coded setting. I'd be prefer to keep the default at 10 for consistency too.

More advanced changes than that are very tricky because Prefect Cloud also uses some of Server's scheduling logic. For bigger changes, you might be better off writing your own scheduling service or using a custom Server version.

**bouke.krom**: OK, clear! I can see how consistency is a key concern too, thanks for the reponse. I'll start working on a PR. Is it customary to make an issue first?
(btw I see someone already tried to get the most minimal fix in: <https://github.com/PrefectHQ/server/pull/225>)

**kevin701**: I don’t think you need to make an issue first, but it would probably be preferable. I asked Nicholas about this and he said there was just no support to push this through. The implementation with the `config.toml` is probably the better approach anyway

**bouke.krom**: <@ULVA73B9P> open "Make max number of scheduled runs per flow configurable"

**kevin701**: oh marvin only works for authorized users. :sweat_smile:. Will do it for you.

**kevin701**: <@ULVA73B9P> open “Make max number of scheduled runs per flow configurable” in server

Original thread can be found [here](https://prefect-community.slack.com/archives/CL09KU1K7/p1630310315179900?thread_ts=1630310315.179900&cid=CL09KU1K7).

bouke-sf:
## Summary
Fixes #280 

Add a configuration parameter to the scheduler that allows to change the hardcoded maximum number of upcoming runs that will be scheduled ahead per flow.

## Importance
Changing this setting is needed when a single flow has more than 10 runs scheduled at exactly the same time (e.g. same cronstring). Currently, the scheduler will schedule 10 of them, the other ones will never be scheduled and fail silently.
The next scheduler update, these flow runs will be considered in the past and will not be scheduled anymore. Adding the configuration option will enable users to resolve the issue -- even if it is still a tricky one to detect.

## Checklist
This PR:

### DIFF
--- a/changes/issue280.yaml
+++ b/changes/issue280.yaml
@@ -1,0 +1,5 @@
+enhancement:
+  - "Make maximum number of scheduled runs per flow configurable - [Issue #280](https://github.com/PrefectHQ/server/issues/280)"
+
+contributor:
+  - "[Bouke Krom](https://github.com/bouke-sf)"
--- a/src/prefect_server/api/flows.py
+++ b/src/prefect_server/api/flows.py
@@ -644,11 +644,9 @@ async def schedule_flow_runs(flow_id: str, max_runs: int = 10) -> List[str]:
         return run_ids
     else:
         # attempt to pull the schedule from the flow group if possible
-        if flow.flow_group.schedule:
-            flow_schedule = flow.flow_group.schedule
         # if not possible, pull the schedule from the flow
-        else:
-            flow_schedule = flow.schedule
+        flow_schedule = flow.flow_group.schedule or flow.schedule
+
         try:
             flow_schedule = schedule_schema.load(flow_schedule)
         except Exception as exc:
--- a/src/prefect_server/api/flows.py
+++ b/src/prefect_server/api/flows.py
@@ -595,7 +595,7 @@ async def set_schedule_inactive(flow_id: str) -> bool:
 
 
 @register_api("flows.schedule_flow_runs")
-async def schedule_flow_runs(flow_id: str, max_runs: int = None) -> List[str]:
+async def schedule_flow_runs(flow_id: str, max_runs: int = 10) -> List[str]:
     """
     Schedule the next `max_runs` runs for this flow. Runs will not be scheduled
     if they are earlier than latest currently-scheduled run that has auto_scheduled = True.
@@ -609,10 +609,6 @@ async def schedule_flow_runs(flow_id: str, max_runs: int = None) -> List[str]:
     Returns:
         - List[str]: the ids of the new runs
     """
-
-    if max_runs is None:
-        max_runs = 10
-
     if flow_id is None:
         raise ValueError("Invalid flow id.")
 
--- a/src/prefect_server/config.toml
+++ b/src/prefect_server/config.toml
@@ -77,6 +77,9 @@ format = "[%(asctime)s] %(levelname)s - %(name)s | %(message)s"
     [services.lazarus]
     resurrection_attempt_limit = 3
 
+    [services.towel]
+    max_scheduled_runs_per_flow = 10
+
 
 # UI Colors corresponding to run states
 [state_colors]
--- a/src/prefect_server/services/towel/scheduler.py
+++ b/src/prefect_server/services/towel/scheduler.py
@@ -3,6 +3,7 @@ import asyncio
 from prefect import api, models
 from prefect.utilities.graphql import EnumValue
 
+from prefect_server import config
 from prefect_server.services.loop_service import LoopService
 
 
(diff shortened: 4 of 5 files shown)


zangell44 (approved):
Looks good to me, thanks for contributing @bouke-sf!!

bouke-sf:
Thanks for reviewing! Fixing this issue is a small price to pay for how useful Prefect is to us, and you guys were very welcoming.

PR #282, opened 2021-09-02
Add loadBalancerIP option for Apollo Service

olwee:
Adds a new option for loadBalancerIP for Apollo Service template, will allow for the use of fixed static public ips when using cloud platforms.

## Summary

Adds a new option for loadBalancerIP for Apollo Service template, will allow for the use of fixed static public ips when using cloud platforms.

## Importance

Without this change, the apollo service IP will be changed after each deployment, requiring end users to keep updating the new API endpoint.

## Checklist

This PR:

### DIFF
--- a/helm/prefect-server/templates/apollo/service.yaml
+++ b/helm/prefect-server/templates/apollo/service.yaml
@@ -14,6 +14,9 @@ spec:
       targetPort: http
       protocol: TCP
       name: http
+  {{- with .Values.apollo.service.publicIP }}
+  loadBalancerIP: {{ . }}
+  {{- end }}
   selector:
     {{- include "prefect-server.matchLabels" . | nindent 4 }}
     app.kubernetes.io/component: apollo


helm/prefect-server/templates/apollo/service.yaml
```diff
@@ -14,6 +14,9 @@ spec:
       targetPort: http
       protocol: TCP
       name: http
+  {{- with .Values.apollo.service.publicIP }}
+  loadBalancerIP: {{ . }}
+  {{- end }}
```
madkinsz, line 19: I think this should just allow extra data to be dumped into this since the type is not always a `LoadBalancer`

If you'd like to contribute this, you'll need to add `apollo.service.extras`  to `values.yaml` and then unpack `extras` here. You can document this `loadBalancerIP` usage in the helm chart readme.

madkinsz (changes requested):

PR #285, opened 2021-09-05
Add debug log for loop services

jlowin:


madkinsz:
Fine with me but we can also pin Cloud to this commit without merging / releasing to Server.

PR #287, opened 2021-09-09
allow query limit override in get_runs_in_queue

jacksund:
## Summary
This extends the graphql endpoint for `get_runs_in_queue`. Rather than being entirely dependent on the configuration kwarg `queued_runs_returned_limit`, it also allows the limit to be overwritten in individual calls.

Because the query already uses the limit tag, this should have no effect on query performance.

## Importance

From discussions with @madkinsz on slack

Some target applications require Agents that query for a fixed number of flows at a time. While this can be set when using Prefect Server config settings, we'd also like to expose this setting to Prefect Agents (and by extension Prefect Cloud users).

As an example use case, other workflow managements softwares (such as [Fireworks](https://github.com/materialsproject/fireworks)) rely on querying/deploying 1 flow at a time in an ad-hoc style. This feature would facilitate transitioning from Fireworks to Prefect and would also enable their use cases.

## Checklist

This PR:

### DIFF
--- a/src/prefect_server/api/runs.py
+++ b/src/prefect_server/api/runs.py
@@ -444,6 +444,7 @@ async def get_runs_in_queue(
     before: datetime = None,
     labels: Iterable[str] = None,
     agent_id: str = None,
+    limit: int = None,
 ) -> List[str]:
 
     if tenant_id is None:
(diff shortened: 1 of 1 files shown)


jacksund:
I haven't added tests for this yet because I'd like to see if this feature would be accepted first.

madkinsz:
Sorry this stagnated, I can barely remember the original internal discussion but I do not think we want to support this pattern in Cloud.

As always, thanks for your contributions jacksund.

jacksund:
No worries! I was able to get around my issue thanks to the new queues / agent api in v2. So this feature isn't necessary anymore anyways. :smile:

PR #289, opened 2021-09-16
Fix image pull secrets in Prefect Server helm chart

### LINKED ISSUE #286: Helm Chart: Invalid format for Kubernetes Agent jobs' ImagePullSecrets
## Description

The helm chart inserts the values for `agent.job.imagePullSecrets` in an incorrect format in the deployment spec, causing the job pods to be unable to connect to private container registries.

### Attempt 1

When this is entered into `values.yaml`:
```yaml
agent:
  job:
    imagePullSecrets:
      - my-secret
```

The `prefect-agent` Deployment has the following environment set; notice the braces and quotes around it.
```yaml
spec:
  template:
    spec:
      containers:
      - env:
        - name: IMAGE_PULL_SECRETS
           value: '["my-secret"]'
```

And when running flows, the generated jobs look like this:
```yaml
spec:
  imagePullSecrets:
  - name: '["my-secret"]'
```

This is not valid, as Kubernetes expects a simple string with the name of the Secret.

### Attempt 2

I also tried to enter a simple string (instead of an array) in `values.yaml`:
```yaml
agent:
  job:
    imagePullSecrets: my-secret
```

Resulting in this in the Deployment's yaml:
```yaml
spec:
  template:
    spec:
      containers:
      - env:
        - name: IMAGE_PULL_SECRETS
           value: '"my-secret"'
```

And in the generated Jobs:
```yaml
spec:
  imagePullSecrets:
  - name: '"my-secret"'
```

Again, there are the extraneous double-quotes round the string.

## Expected Behavior

The generated YAML should look like this for both the agent Deployment and generated Jobs.

```yaml
spec:
  imagePullSecrets:
  - name: my-secret
```

## Reproduction
See above.

The cause of this seems to be a mismatch between the Helm chart template that inserts the ImagePullSecrets as a JSON string:

https://github.com/PrefectHQ/server/blob/673cbeab144cb2ae06a17c23fdfb4e2356f0376c/helm/prefect-server/templates/agent/deployment.yaml#L50-L51

And the agent, which expects the environment variable to be a simple comma-delimited string (not JSON):

https://github.com/PrefectHQ/prefect/blob/5221e6c4ef68eb8c7659c212ea937d1856cee8a4/src/prefect/agent/kubernetes/agent.py#L114-L120

## Environment
Kubernetes: Azure Kubernetes Services
Prefect Server Helm chart version: 2021.07.06

mashun4ek:
## Summary

In aim to connect to private container registries, you can provide k8s secrets to `imagePullSecrets` on Prefect agent deployment. You can set env var `IMAGE_PULL_SECRETS` on `prefect-agent` deployment, or if using helm chart, you can set 
```
agent:
  job:
    imagePullSecrets:
      - secret_name1
      - secret_name2
```
Instead of modifying the logic [here](https://github.com/PrefectHQ/prefect/blob/2cd13929c7701e21b8b865c36291bdee39209b95/src/prefect/agent/kubernetes/agent.py#L115), let's use Go template function `join` to convert list to a string.

## Importance

Closes #286 

@TylerWanner can you please take a look?
## Checklist

This PR:

- [-] adds new tests (if appropriate)

### DIFF
--- a/changes/pr289.yaml
+++ b/changes/pr289.yaml
@@ -1,0 +1,2 @@
+fix:
+  - "Fix setting `imagePullSecrets` in k8s jobs with Prefect flows in Prefect Server Helm chart - [#289](https://github.com/PrefectHQ/server/pull/289)"
--- a/helm/prefect-server/templates/agent/deployment.yaml
+++ b/helm/prefect-server/templates/agent/deployment.yaml
@@ -24,7 +24,13 @@ spec:
         {{- include "prefect-server.matchLabels" . | nindent 8 }}
         app.kubernetes.io/component: agent
     spec:
-      imagePullSecrets: {{ concat .Values.agent.image.pullSecrets .Values.imagePullSecrets | uniq | toJson }}
+    {{- if .Values.agent.image.pullSecrets }}
+      imagePullSecrets:
+        {{- toYaml .Values.agent.image.pullSecrets | nindent 8 }}
+    {{- else if .Values.imagePullSecrets }}
+      imagePullSecrets:
+        {{- toYaml .Values.imagePullSecrets | nindent 8 }}
+    {{- end }}
       serviceAccountName: {{ include "prefect-server.serviceAccountName" . }}
       {{- with .Values.agent.podSecurityContext }}
       securityContext:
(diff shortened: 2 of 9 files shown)


madkinsz:
@TylerWanner can you review?

TylerWanner (approved):

madkinsz (approved):

PR #291, opened 2021-09-20
submitted state lock

### LINKED ISSUE #203: Add submitted state locking to Prefect Server
If a user has multiple agents with the same set of labels, multiple agents can pick up and start the same flow run. When running with Prefect Cloud, this will be caught via submitted state locking, so only one flow run can progress (and the others will abort). This feature was never ported over to Prefect Server, so currently users can end up with multiple executions of the same flow run.

jakekaplan:
## Summary
Adds an in memory in process submitted state lock. Closes #203

## Importance
Currently it is possible for multiple agents to pick up the same scheduled flow_run, this enables a buffer so that if multiple agents try to run the same flow_run only one will succeed. 

## Checklist

This PR:

### DIFF
--- a/src/prefect_server/utilities/collections.py
+++ b/src/prefect_server/utilities/collections.py
@@ -1,5 +1,6 @@
 import itertools
-from typing import Iterable
+from typing import Hashable, Iterable, Optional
+import threading
 
 
 def chunked_iterable(iterable: Iterable, size: int):
@@ -19,3 +20,60 @@ def chunked_iterable(iterable: Iterable, size: int):
         if not chunk:
             break
         yield chunk
+
+
+class TimedUniqueValueStore:
+    """
+    Stores unique values for a specific amount of time.
+
+    Args:
+        - duration (float): the number of seconds to store a value before removing it
+    """
+    def __init__(self, duration: float = 10):
+        self.values = set()
+        self.lock = threading.Lock()
+        self.duration = duration
+
+    def add_and_expire(self, value: Hashable, duration: Optional[float] = None) -> bool:
+        """
+        Add a unique value to store and kick off a timer to remove it
+
+        Args:
+            - value (Hashable): value to store, must be hashable and
+                able to be stored in a set
+            - duration (float, optional): the number of seconds to store a value
+                before removing it. Defaults to self.duration if not provided
+        Returns:
+            - bool: `True` if the value did not already exist in the store
+                and is successfully added, otherwise `False`
+        """
+        if duration is None:
+            duration = self.duration
+
+        with self.lock:
+            if value in self.values:
+                return False
+            else:
+                self.values.add(value)
+                threading.Timer(duration, self.remove, kwargs={"value": value}).start()
+                return True
+
+    def remove(self, value: Hashable):
+        """
+        Remove a value from the store
+
+        Args:
+            - value (Hashable): value to remove from store
+        """
+        with self.lock:
+            self.values.discard(value)
+
+    def exists(self, value: Hashable):
+        """
+        Check if a value exists in the store
+
+        Args:
+            - value (Hashable): value to remove from store
+        """
+        with self.lock:
+            return value in self.values
--- a/src/prefect_server/utilities/collections.py
+++ b/src/prefect_server/utilities/collections.py
@@ -22,59 +22,63 @@ def chunked_iterable(iterable: Iterable, size: int):
         yield chunk
 
 
-class TimedUniqueValueStore:
+class ExpiringSet:
     """
-    Stores unique values for a specific amount of time.
+    Stores values in a set for a specific amount of time.
 
     Args:
-        - duration (float, int): the number of seconds to store a value before removing it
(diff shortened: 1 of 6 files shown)


src/prefect_server/utilities/collections.py
```diff
@@ -19,3 +20,61 @@ def chunked_iterable(iterable: Iterable, size: int):
...
+            - duration (float, int, optional): the number of seconds to store a value
+                before removing it. Defaults to self.duration if not provided
+        Returns:
+            - bool: `True` if the value did not already exist in the store
+                and is successfully added, otherwise `False`
+        """
+        if duration is None:
+            duration = self.duration
+
+        with self.lock:
```
cicdw, line 54: Are there any edge cases where we may want to put on a timeout on the lock acquisition?  E.g., instead of using the context manager do something like:
```python
acquired = self.lock.acquire(timeout=2)
try:
    if not acquired:
        return False
    ... # value check, etc.
finally:
    self.lock.release()

return bool_val
```

the downside ofc is that it's more complicated

src/prefect_server/utilities/collections.py
```diff
@@ -19,3 +20,61 @@ def chunked_iterable(iterable: Iterable, size: int):
...
+            if value in self.values:
+                return False
+            else:
+                self.values.add(value)
+                threading.Timer(duration, self.remove, kwargs={"value": value}).start()
```
cicdw, line 59: clever!

src/prefect_server/utilities/collections.py
cicdw: We could write a custom context manager that handles this and use it instead of the stdlib one (if we think it's a good idea)

src/prefect_server/utilities/collections.py
```diff
@@ -19,3 +20,61 @@ def chunked_iterable(iterable: Iterable, size: int):
...
+        """
+        Remove a value from the store
+
+        Args:
+            - value (Hashable): value to remove from store
+        """
+        with self.lock:
+            self.values.discard(value)
+
+    def exists(self, value: Hashable):
```
cicdw, line 72: Looks like this method didn't get used - intentional or oversight?

cicdw (commented):
A question about timeouts for acquiring the lock; also if you include "Closes https://github.com/PrefectHQ/server/issues/203" in the main PR description GitHub will auto-close the issue 

Otherwise this looks clean and robust, nice work!

src/prefect_server/utilities/collections.py
```diff
@@ -19,3 +20,61 @@ def chunked_iterable(iterable: Iterable, size: int):
...
             break
         yield chunk
+
+
+class TimedUniqueValueStore:
+    """
+    Stores unique values for a specific amount of time.
+
+    Args:
+        - duration (float, int): the number of seconds to store a value before removing it
```
madkinsz, line 30: Usually we try to include the unit so it's clear when you call it e.g. `duration_seconds`

src/prefect_server/utilities/collections.py
```diff
@@ -19,3 +20,61 @@ def chunked_iterable(iterable: Iterable, size: int):
         if not chunk:
             break
         yield chunk
+
+
+class TimedUniqueValueStore:
```
madkinsz, line 25: Minor nit: Why not just `ExpiringSet` or similar?

src/prefect_server/utilities/collections.py
```diff
@@ -19,3 +20,61 @@ def chunked_iterable(iterable: Iterable, size: int):
...
+    Args:
+        - duration (float, int): the number of seconds to store a value before removing it
+    """
+
+    def __init__(self, duration: Union[float, int] = 10):
+        self.values = set()
+        self.lock = threading.Lock()
+        self.duration = duration
+
+    def add_and_expire(self, value: Hashable, duration: Optional[float, int] = None) -> bool:
```
madkinsz, line 38: It doesn't seem like we actually need the `duration` override here for any of the current use-cases. can we just call this `add`?

src/prefect_server/utilities/collections.py
jakekaplan: My naming ability is a frankly an evolving skill. I like that a lot better

src/prefect_server/utilities/collections.py
jakekaplan: yes can do 👍

src/prefect_server/utilities/collections.py
madkinsz: Unless we have a concrete case, I don't think we need to do this. I've seen lots of usage of the stdlib context manager without timeouts.

jakekaplan: I'll leave it without for now, but can always go back and add

cicdw (commented):
@jakekaplan this LGTM but would you also mind adding a test at the GraphQL layer within `tests/graphql/test_states.py` that replicates the API level test at a higher level ?

jakekaplan:
@cicdw added the graphql test 👍

cicdw (approved):

PR #298, opened 2021-10-06
Add flow defaults to flow runs

jakekaplan:
## Summary
Adds the default parameters from the flow to the flow run. This 

## Importance

Resolves https://github.com/PrefectHQ/prefect/issues/5009#issuecomment-934520626.
This will allow for flow run parameters to appear consistent whether they are set from the flow, the flow run group or the flow run itself.

## Checklist

This PR:

### DIFF
--- a/src/prefect_server/api/runs.py
+++ b/src/prefect_server/api/runs.py
@@ -160,9 +160,10 @@ async def create_flow_run(
             run_config = flow.run_config
 
     # check parameters
-    run_parameters = {param["name"]: param["default"] for param in flow.parameters if param["default"]}
+    run_parameters = {
+        param["name"]: param["default"] for param in flow.parameters if param["default"]
+    }
     run_parameters.update(flow.flow_group.default_parameters or {})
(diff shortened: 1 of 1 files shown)


src/prefect_server/api/runs.py
```diff
@@ -160,7 +160,10 @@ async def create_flow_run(
             run_config = flow.run_config
 
     # check parameters
-    run_parameters = flow.flow_group.default_parameters
+    run_parameters = {
+        param["name"]: param["default"] for param in flow.parameters if param["default"]
+    }
+    run_parameters.update(flow.flow_group.default_parameters or {})
```
zangell44, line 166: should the flow group parameters take precedence over the flow parameters here? it seems like the flow's params would take priority

zangell44: 👍 I'd preserve whatever order is there today

zangell44: okay yep I see that order in a test case, lgtm

PR #317, opened 2021-11-23
Prevent create_flow_run from succeeding with bad params

pjanowski:
## Summary
Fix `create_flow_run` to disallow incorrect flow parameters.

## Importance
This PR fixes the `runs.create_flow_run` mutation to disallow passing of bad parameters. Currently, if a user invokes `create_flow_run` and passes parameters that do not exist in the flow for which a run is being created, the flow run will be created successfully with the bad parameters. These will be noop's and the flow will run with it's default parameter values. (eg. if a flow run will has parameter 'pizza' with default value 'pepperoni' and the user tries to create a flow run with parameter 'piza' and value 'sausage', the flow run will successfully be created with parameters `{'pizza':'pepperoni', 'piza':'sausage'}` ). This leads to unintended and hard to debug behavior. The fix checks to make sure that all the parameters passed by the caller are valid parameters of the flow being invoked. 

## Checklist

This PR:

### DIFF
--- a/src/prefect_server/api/runs.py
+++ b/src/prefect_server/api/runs.py
@@ -164,18 +164,19 @@ async def create_flow_run(
         param["name"]: param["default"] for param in flow.parameters if param["default"]
     }
 
-    # check for supplied params not in run params
-    if parameters:
-        unknown_parameters = set(parameters.keys()).difference(
-            set(run_parameters.keys())
-        )
-        if unknown_parameters:
-            raise ValueError(
-                f"Unknown parameters supplied to Flow {flow.id}'s. Unknown parameters: {unknown_parameters}."
-            )
-
     run_parameters.update(flow.flow_group.default_parameters or {})
     run_parameters.update((parameters or {}))
+
+    # check for supplied params not in the flow's params
+    unknown_parameters = set(run_parameters.keys()).difference(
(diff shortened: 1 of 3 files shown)


src/prefect_server/api/runs.py
```diff
@@ -163,6 +163,17 @@ async def create_flow_run(
     run_parameters = {
         param["name"]: param["default"] for param in flow.parameters if param["default"]
     }
+
+    # check for supplied params not in run params
```
zangell44, line 167: I think this check should occur after the next line so flow group parameters can also be supplied (`run_parameters.update(flow.flow_group.default_parameters or {})`)

zangell44: also this is only checking against flow parameters that have default values

pjanowski: Thanks for the good catch! Please see now.

pjanowski:
Closing because it's been a while and I think Prefect team has better ideas for Orion.

madkinsz:
We were thinking about incorporating this as a backwards compatible change but didn't want to burden you with toggling the behavior based on client version. We were also not sure if we should just make the breaking change since it was bad behavior in the first place.

I appreciate your good attitude :) we still may incorporate this fix.

PR #321, opened 2021-11-26
Clarify how to set password for the Postgres database in the Server Helm chart documentation

anna-geller:
Clarifying how the Kubernetes secret should be set on the Helm chart values when using an existing Postgres database rather than the one included with a sub-chart.

### DIFF
--- a/helm/prefect-server/README.md
+++ b/helm/prefect-server/README.md
@@ -197,6 +197,16 @@ We strongly recommend that you do not deploy a production database using this ch
 The provided database will **not** persist your data by default.
 When connecting to an external database, PostgreSQL 11+ is recommended.
 
+In order to use an external database with this Helm chart, you need to create a Kubernetes secret that will contain the database password. This secret must be a key-value pair containing the key `postgresql-password`. You need to create this secret yourself before referencing the secret name under `existingSecret`. The name of the Kubernetes secret is arbitrary, e.g.: 
+
+    kubectl create secret generic prefect-postgresql-pwd --from-literal='postgresql-password=YOUR_POSTGRES_PWD'
+
+Here is how you could use this secret in values.yaml:
+
+    existingSecret: prefect-postgresql-pwd
+
+Note that the argument `postgresqlPassword` will be ignored in this case. This argument is only relevant when using the Postgres database included in the chart. For an external database, you must create and use `existingSecret` instead of `postgresqlPassword`.
+
 An external database will require some minimal setup for Hasura.
 The following needs to be run or the user should have permissions to execute it and Hasura will run it on startup:
 
--- a/helm/prefect-server/README.md
+++ b/helm/prefect-server/README.md
@@ -197,13 +197,13 @@ We strongly recommend that you do not deploy a production database using this ch
 The provided database will **not** persist your data by default.
 When connecting to an external database, PostgreSQL 11+ is recommended.
 
-In order to use an external database with this Helm chart, you need to create a Kubernetes secret that will contain the database password. This secret must be a key-value pair containing the key `postgresql-password`. You need to create this secret yourself before referencing the secret name under `existingSecret`. The name of the Kubernetes secret is arbitrary, e.g.: 
+In order to use an external database with this Helm chart, you need to create a Kubernetes secret that will contain the database password. This secret must be a key-value pair containing the key `postgresql-password`. You need to create this secret yourself, then reference the secret name using `existingSecret`. The name of the Kubernetes secret is arbitrary, e.g.: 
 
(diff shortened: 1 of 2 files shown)


madkinsz (changes requested):
Can you also update the PR title to capture the intent?

helm/prefect-server/values.yaml
```diff
@@ -52,8 +52,15 @@ postgresql:
...
-  # Kubernetes secret containing the key `postgresql-password`
+  # postgres service, this value should be explicitly set to
+  # the name of a Kubernetes secret. This secret must be a
+  # key-value pair containing the key `postgresql-password`.
+  # Note that you need to create this secret yourself before
+  # referencing the secret name under `existingSecret`.
+  # The name of the Kubernetes secret is arbitrary.  e.g.:
+  # kubectl create secret generic prefect-postgresql-pwd
+  # --from-literal='postgresql-password=YOUR_POSTGRES_PWD'
+  # Helm chart value: existingSecret: prefect-postgresql-pwd
```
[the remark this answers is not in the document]
anna-geller: That makes sense, thanks a lot!

helm/prefect-server/values.yaml
```diff
@@ -52,14 +52,20 @@ postgresql:
   # existingSecret configures which secret should be referenced
   # for access to the database. If null and `useSubChart` is
   # enabled, the secret will be generated. If using an external
-  # postgres service, this value should be set to the name of a
-  # Kubernetes secret containing the key `postgresql-password`
+  # postgres service, this value should be set to the name of 
+  # an existing Kubernetes secret. This secret must contain 
+  # a key-value pair where the key is postgresql-password
```
madkinsz, line 57:
```suggestion
  # a key-value pair where the key is `postgresql-password `
```

helm/prefect-server/README.md
```diff
@@ -197,6 +197,16 @@ We strongly recommend that you do not deploy a production database using this ch
...
 When connecting to an external database, PostgreSQL 11+ is recommended.
 
+In order to use an external database with this Helm chart, you need to create a Kubernetes secret that will contain the database password. This secret must be a key-value pair containing the key `postgresql-password`. You need to create this secret yourself before referencing the secret name under `existingSecret`. The name of the Kubernetes secret is arbitrary, e.g.: 
+
+    kubectl create secret generic prefect-postgresql-pwd --from-literal='postgresql-password=YOUR_POSTGRES_PWD'
+
+Here is how you could use this secret in values.yaml:
+
+    existingSecret: prefect-postgresql-pwd
+
```
madkinsz, line 207: Since this is not self contained, I'd rather suggest installation with this key as the above examples do

> Then, add the flag `--set postgresql.existingSecret=prefect-postgresql-pwd` to the `helm install` command to use this secret to connect to your database.

helm/prefect-server/README.md
```diff
@@ -197,6 +197,16 @@ We strongly recommend that you do not deploy a production database using this ch
 The provided database will **not** persist your data by default.
 When connecting to an external database, PostgreSQL 11+ is recommended.
 
+In order to use an external database with this Helm chart, you need to create a Kubernetes secret that will contain the database password. This secret must be a key-value pair containing the key `postgresql-password`. You need to create this secret yourself before referencing the secret name under `existingSecret`. The name of the Kubernetes secret is arbitrary, e.g.: 
+
+    kubectl create secret generic prefect-postgresql-pwd --from-literal='postgresql-password=YOUR_POSTGRES_PWD'
```
madkinsz, line 202: Can you put this in code blocks?

helm/prefect-server/README.md
```diff
@@ -197,6 +197,16 @@ We strongly recommend that you do not deploy a production database using this ch
 The provided database will **not** persist your data by default.
 When connecting to an external database, PostgreSQL 11+ is recommended.
 
+In order to use an external database with this Helm chart, you need to create a Kubernetes secret that will contain the database password. This secret must be a key-value pair containing the key `postgresql-password`. You need to create this secret yourself before referencing the secret name under `existingSecret`. The name of the Kubernetes secret is arbitrary, e.g.:
```
madkinsz, line 200:
```suggestion
In order to use an external database with this Helm chart, you need to create a Kubernetes secret that will contain the database password. This secret must be a key-value pair containing the key `postgresql-password`. You need to create this secret yourself then reference the secret name using `existingSecret`. The name of the Kubernetes secret is arbitrary, e.g.: 
```

helm/prefect-server/README.md
anna-geller: yup, check 👍

madkinsz (approved):

PR #322, opened 2021-11-29
[Feature] Allows custom job templates on helm chart

gabriel-milan:
## Summary
This allows setting a custom job template directly on the `values.yaml` file.

## Importance
In our use case, the only thing we needed to do was to set environment variables for all our jobs using `envFrom`. With this change (tested and working for us) it allows us to change the job template directly on the `values.yaml` file, making it much easier to manage.

## Checklist

This PR:

- [x] adds new tests (not applicable, I guess. please correct me if I'm wrong)
- [x] adds a change file in the `changes/` directory (wip)

### DIFF
--- a/changes/pr322.yaml
+++ b/changes/pr322.yaml
@@ -1,0 +1,23 @@
+# An example changelog entry
+#
+# 1. Choose one (or more if a PR encompasses multiple changes) of the following headers:
+#   - feature
+#   - enhancement
+#   - fix
+#   - deprecation
+#   - breaking (for breaking changes)
+#   - migration (for database migrations)
+#
+# 2. Fill in one (or more) bullet points under the heading, describing the change.
+#    Markdown syntax may be used.
+#
+# 3. If you would like to be credited as helping with this release, add a
+#    contributor section with your name and github username.
+#
+# Here's an example of a PR that adds an enhancement
+
+feature:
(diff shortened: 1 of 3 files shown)


madkinsz:
Can you share an example of using this? It seems like the user experience would not really be great from a helm-chart perspective as you'd have to ensure the file is available.

gabriel-milan:
Sure! In our deployment, we're setting `agent.jobTemplate` to a YAML hosted on GCS. The YAML looks like this:

```yaml
apiVersion: batch/v1
kind: Job
spec:
  template:
    spec:
      containers:
        - name: flow
          envFrom:
            - secretRef:
                name: gcp-credentials
```

This allows us to set many environment variables from existing secrets in our different namespaces, which eases management of dev/staging/prod environments.

madkinsz:
Perhaps we should name it "jobTemplateFilePath" or something instead in `values.yaml`. I expect to be able to inline some yaml as currently named.

gabriel-milan:
That makes sense

madkinsz:
https://github.com/PrefectHQ/server/pull/323 will resolve the CI failures

madkinsz:
@gabriel-milan I don't have permission to update your branch

gabriel-milan:
That's weird. I've enabled the `Allow edits by maintainers`. Is there anything else I can do?

madkinsz:
@gabriel-milan that works! If this passes CI I'll get it into #324

madkinsz (approved):

PR #328, opened 2021-12-14
upgrade to hasura2.0

jakekaplan:
## Summary

Upgrades Hasura to v.2.0.9

## Importance
Prefect Server is on a very old version of Hasura. Upgrading to v.2.0.9 allows it to keep up with current functionality.

## Checklist

This PR:

~~- [ ] adds new tests (if appropriate)~~

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -63,10 +63,11 @@ executors:
           POSTGRES_DB: prefect_server
         command: postgres -c max_connections=150
 
-      - image: "hasura/graphql-engine:v1.3.3"
+      - image: "hasura/graphql-engine:v2.0.9"
         command: /bin/sh -c "sleep 5 && graphql-engine serve"
         environment:
           HASURA_GRAPHQL_DATABASE_URL: postgres://prefect:prefect@localhost:5432/prefect_server
+          HASURA_GRAPHQL_V1_BOOLEAN_NULL_COLLAPSE: "true"
           HASURA_GRAPHQL_ENABLE_CONSOLE: "true"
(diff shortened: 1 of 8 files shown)


cicdw:
This LGTM -- we just need to confirm that the metadata that Hasura manages within the DB isn't negatively affected by this change; we'll need to reference the section titled "Deprecation of database specific env vars" in [the upgrade guide](https://hasura.io/docs/latest/graphql/core/guides/upgrade-hasura-v2.html).

I think one way of approaching this is to start Server connected to a persistent database with the older version of Hasura, and then do the Hasura upgrade only (with the same DB) and confirm all works as expected

jakekaplan:
@cicdw I tested this with a persistent postgres on my local machine. Hasura will manage the hdb_catalog schema itself. It effectively removes the tables used for v1 Hasura and populates v2 Hasura tables. Really what it does it shoves all that table information into a single json blob in a `hdb_metadata` table. Which exactly you get out when you export the metadata table via the UI

jakekaplan:
@cicdw I updated the yaml to "hasura config verion 3" to keep it consistent with cloud. We don't actually set anything different here (like max_connections) in cloud, but this allows for it if we want to.

BitTheByte:
@cicdw this change seems ok, I tested the server's functionality using my local environment and everything apparently is working fine. I'd love to see it moves to release as I'm also using Hasura for my deployment and it would be easier to just have one instance instead of two

cicdw:
Thanks for testing @BitTheByte , we really appreciate that!

cicdw (approved):
LGTM

Issue #331, opened 2022-01-07
Helm default values for env are of incorrect type


oreonl:
## Description
The default values for the different env variables in the helm chart values.yaml file are set to an empty map, but the templates are setup so that you need to pass in a list. This causes a conflict when you try and set these values.

## Expected Behavior
We could set the env values to add additional environment variables on the containers.

## Reproduction
helm template -s templates/graphql/deployment.yaml --set graphql.env[0].value=0.15.11  --set graphql.env[0].name=PREFECT_CORE_VERSION --set postgresql.useSubChart=false . -f values.yaml

Error: failed parsing --set data: unable to parse key: interface conversion: interface {} is map[string]interface {}, not []interface {}

## Environment
<!-- Any additional information about your environment

oreonl:
Fixed under https://github.com/PrefectHQ/server/pull/333

PR #333, opened 2022-01-10
fix: Change default helm env values from map to list

oreonl:
## Summary

The default values for the different env variables in the helm chart values.yaml file are set to an empty map, but the templates are setup so that you need to pass in a list. This causes a conflict when you try and set these values. This PR updates the default env values to an empty list.

See issue https://github.com/PrefectHQ/server/issues/331

## Importance

If you attempt to pass in values for the env you get an error:
```
Error: failed parsing --set data: unable to parse key: interface conversion: interface {} is map[string]interface {}, not []interface {}
```

## Checklist

This PR:

### DIFF
--- a/changes/issue331.yaml
+++ b/changes/issue331.yaml
@@ -1,0 +1,5 @@
+fix:
+  - "Fix default env values for helm chart - [#331](https://github.com/PrefectHQ/server/issues/331)"
+
+contributor:
+  - "[Oreon Lothamer](https://github.com/oreonl)"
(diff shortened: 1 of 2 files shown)


madkinsz (approved):
Thanks!

madkinsz (commented):
Actually, would it make sense to adjust the templating instead? Could you provide an example env value for `values.yaml` now?

oreonl:
Here is an example value
```
graphql:
  env:
   - name: PREFECT_CORE_VERSION
     value: 0.15.11
```

madkinsz:
Ah yeah that is fine with me. Thanks again!

PR #276, opened 2021-08-25
change RoleBinding API to rbac.authorization.k8s.io/v1

gmaliar:
## Summary
Warning when applying the helm charts 

rbac.authorization.k8s.io/v1beta1 RoleBinding is deprecated in v1.17+, unavailable in v1.22+; use rbac.authorization.k8s.io/v1 RoleBinding

## Importance
It would break the helm chart for anybody who is using v1.22+
Kubernetes 1.17 support has been dropped with the release of v1.22

## Checklist

This PR:

- [ ] adds new tests (if appropriate)
- [ ] adds a change file in the `changes/` directory (if appropriate)

PR #345, opened 2022-02-23
Update ingress for the ui and apollo from v1beta1 to v1

flakeman2:
## Summary
With the latest versions of Kubernetes and Ingress Nginx, networking.k8s.io has moved from v1beta1 to v1.  This PR is to reflect that change.  Also, the yaml has been updated to work with networking.k8s.io/v1

https://github.com/PrefectHQ/server/pull/276 -- this was a recent and similar change 

## Importance
Without this change the helm chart will fail for anyone using v1.22+
Kubernetes 1.17 support has been dropped with the release of v1.22

## Checklist

This PR:


madkinsz:
Thanks! This LGTM. I'll need to find some time to sort out these CI issues before merging though.

PR #346, opened 2022-02-24
fix missing server configs on hasura and towel deployments

gabriel-milan:
## Summary
As I've spoken to @kvnkho [on Slack](https://prefect-community.slack.com/archives/CL09KU1K7/p1644943315764649), the Prefect Server configurations set on the `values.yaml` file are still missing on both Hasura and Towel deployments.

## Importance
This fixes misconfiguration for Hasura and Towel.

## Checklist

This PR:


madkinsz:
Hi! We can add this to the towel service, which indeed does consume some of the config values for Prefect Server. However, Hasura does not use our config so I don't see what including the environment variables there would do. Is there a specific issue that's resolving for you?

gabriel-milan:
That's correct. Only Towel needs the update for my use case. I just thought that maybe it was also missing for Hasura, but if it's not used, I think that it's fine.

madkinsz:
> That's correct. Only Towel needs the update for my use case. I just thought that maybe it was also missing for Hasura, but if it's not used, I think that it's fine.

Yeah their container won't do anything with those variables so we probably shouldn't include them.

gabriel-milan:
> we probably shouldn't include them

I've just removed it, thanks for saying that!

madkinsz:
@gabriel-milan thanks! Looks like you'll just need to get up to date with the `master` branch now.

gabriel-milan:
does it seem correct? i'm not sure

madkinsz (approved):
Yep! Thanks!

PR #366, opened 2022-05-04
Helm: Remove duplicate envvar in graphql deployment

julienduchesne:
This envvar is already defined here: https://github.com/PrefectHQ/server/blob/master/helm/prefect-server/templates/_helpers.tpl#L199-L200. So, it ends up twice in the env map. This fails on a server-side apply


madkinsz (approved):
Thanks!

PR #367, opened 2022-05-14
Chore: Change Zombie Killer log message

Laerte:
This PR closes:
- https://github.com/PrefectHQ/prefect/issues/5717

### DIFF
--- a/src/prefect_server/services/towel/zombie_killer.py
+++ b/src/prefect_server/services/towel/zombie_killer.py
@@ -138,10 +138,11 @@ class ZombieKiller(LoopService):
                 "id": True,
                 "flow_run_id": True,
                 "tenant_id": True,
+                "name": True,
                 # Information about the current flow run state
                 "flow_run": {"state"},
                 # get information about retries from task
-                "task": {"max_retries", "retry_delay", "id", "name"},
+                "task": {"max_retries", "retry_delay"},
                 # count the number of retrying states for this task run
                 with_args(
                     "retry_count: states_aggregate",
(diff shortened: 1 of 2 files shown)


src/prefect_server/services/towel/zombie_killer.py
```diff
@@ -180,7 +180,10 @@ async def reap_zombie_task_runs(
 
                 # mark failed
                 else:
-                    message = "No heartbeat detected from the remote task; marking the run as failed."
+                    message = (
+                        "No heartbeat detected from the remote task; marking the "
+                        f"task run {tr.task.id} ({tr.task.name}) as failed."
```
cicdw, line 185: This logs the task ID and name, not the task run ID and name; I think you meant:
```suggestion
                        f"task run {tr.id} ({tr.name}) as failed."
```
in which case the expanded query above isn't necessary

cicdw (commented):
left a comment about the logged data

src/prefect_server/services/towel/zombie_killer.py
Laerte: @cicdw Thanks! I added name to query because didn't have. I runned tests locally is normal task run name be null?

cicdw: Yes that is normal - task runs only have names when a user explicitly provides them.

Laerte: Great, thanks!

Laerte: Thanks @cicdw, I lookup again the issue and noticed that name is indeed task name not task run name, should i change the PR acordelly?

Laerte:
@cicdw  @madkinsz @zangell44 Can you guys take a look to check if i'm missing something? Thanks.

zangell44 (approved):

PR #370, opened 2022-05-18
Overwrite security context and disable graphql init container

aanogueira:
## Summary

Added possibility to overwrite security context and to disable init container for graphql db upgrades.

## Importance

When using Istio, having the init container causes the the application to malfunction with Istio sidecar injection.
If we update the chart, we don't need the database to always be upgrading. So either adding the possibility to overwrite the init container security context (Istio suggestion `Set the uid of the init container to 1337 using runAsUser. 1337 is the uid used by the sidecar proxy. Traffic sent by this uid is not captured by the Istio's iptables rule. Application container traffic will still be captured as usual.`) or to completely disable would be pretty helpful.

## Checklist

This PR:


madkinsz (commented):
Thanks for contributing! This looks good to me. I'll probably get a review from someone on our platform team as well.

Can you restore the `changes/EXAMPLE.yaml` file?

gabcoyne:
This looks good to me as well

madkinsz (approved):

Issue #374, opened 2022-07-13
Helm dep up failed because Postgresql version


amitza:
## Description
In order to install the prefect helm chart, first all the helm dependencies need to be installed.
When trying to update dependencies the command `helm dep up` fails with the error:
`Error: can't get a valid version for repositories postgresql. Try changing the version constraint in Chart.yaml`

when checking the [bitnami chart repo](https://charts.bitnami.com/bitnami) the only postgresql chart is in version 11.6.16

## Expected Behavior
`helm dep up` installs the dependencies

## Reproduction
Clone this repository and run `helm dep up` in the helm/prefect-server folder

## Environment
helm version: `version.BuildInfo{Version:"v3.3.0", GitCommit:"8a4aeec08d67a7b84472007529e8097ec3742105", GitTreeState:"dirty", GoVersion:"go1.14.7"}`

PR #375, opened 2022-07-14
Update Postgresql chart's version from ~9 to ~11

amitza:
## Summary
Updating the dependencies of the main prefect server chart in order to fix https://github.com/PrefectHQ/server/issues/374

## Importance
The Postgres' current version (9.3.2) isn't accessible through the [Bitnami chart repo](https://charts.bitnami.com/bitnami), the only version there is 11+.

## Checklist

This PR:


madkinsz:
Hi! I see a similar fix at https://github.com/PrefectHQ/server/pull/376 but yours includes some additional changes. Could you explain those?

jawnsy:
Thanks so much for your contribution!

Please make sure that you test the case of upgrading (installing an old version of the chart, upgrading to this one).

I think changing PostgreSQL versions may cause problems when upgrading, because I think there's a manual upgrade step required at the PostgreSQL level - typically taking a backup and restoring: https://www.postgresql.org/docs/14/upgrading.html

The chart also has some documentation describing an upgrade process: https://docs.bitnami.com/kubernetes/infrastructure/postgresql/administration/upgrade/

amitza:
I managed to update the postgresql chart and helm-install it on my KinD cluster.
I'm not sure why the tests fail.

.circleci/config.yml
```diff
@@ -58,21 +58,22 @@ executors:
           PREFECT_SERVER__DATABASE__USERNAME: prefect
           PREFECT_SERVER__DATABASE__PASSWORD: pr343ct
 
-      - image: "postgres:11"
+      - image: "bitnami/postgresql:14.4.0-debian-11-r9"
```
madkinsz, line 61: Hi! Curious why you made changes to the CircleCI config

madkinsz:
@amitza I do not have permissions to push to your branch so I cannot update this with the CI fixes in `master`. 

I'm willing to accept this if you revert your changes to the CircleCI config.

.circleci/config.yml
amitza: I've reverted the postgres version

madkinsz: Thanks! You may need to revert the other changes as well.

If you merge update with `master` this should be able to pass CI.

madkinsz:
Thanks for contributing! This is complete in #384

PR #376, opened 2022-07-14
fix(helm): update postgres dependency version

joelluijmes:
## Summary

Helm would fail error: "Error: can't get a valid version for repositories postgresql. Try changing the version constraint in Chart.yaml". This is because bitnami implemented a retention policy and thereby remove old versions (more info: https://github.com/bitnami/charts/issues/10539). Note, this issue even applies when condition evaluates false.

## Importance

By updating the version to (current) latest chart, helm can now deploy again.

## Checklist

This PR:


madkinsz:
Thanks for the fix! Does this bump the actual postgres version being used or just the chart?

PR #385, opened 2022-09-02
Improve Apollo error formatting to enable client retries and obfuscate IP addresses

madkinsz:
Updates Apollo error formatting to include Prefect Cloud-like behavior in which:
- IP addresses are obfuscated
- API_ERROR is added as an error code for internal errors to signal that the client should retry

Replaces #373 
Paired with https://github.com/PrefectHQ/cloud/pull/4095
Closes https://github.com/PrefectHQ/prefect/issues/5829 for Server users

### DIFF
--- a/services/apollo/src/index.js
+++ b/services/apollo/src/index.js
@@ -25,6 +25,35 @@ const TELEMETRY_ID = uuidv4()
 // Server
 const app = express()
 const depthLimit = require('graphql-depth-limit')
+
+
+/*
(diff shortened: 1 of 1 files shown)


cicdw (approved):

madkinsz:
I want to attempt to do some manual confirmation that this behaves correctly before merging :)

madkinsz:
The first iteration of this was missing some constants but still passed CI which makes me concerned about test coverage here. Leaning away from merging this until we get additional feedback from usage in a real environment.

PR #407, opened 2022-12-06
docs: fix postgresql-password -> postgres-password

joelluijmes:
## Summary

The documentation refers to different secret than is actually used in the Helm chart. [Although, I actually think the secret is renamed at some point 🤷 ].

## Importance

If the wrong secret is configured, the pods will never come online:
```
CreateContainerConfigError: couldn't find key postgres-password in Secret prefect/prefect-postgresql 
```

## Checklist

This PR:


madkinsz:
Thanks for opening the PR! I believe this was introduced in https://github.com/PrefectHQ/server/pull/384

madkinsz (approved):

